PR #23, opened 2015-06-01
Add screenshot to README.

epost:


rikvdkleij:
@epost It's not (yet) possible to compile Haskell files from within IntelliJ, so maybe the message in status bar is misleading?

epost:
Hm, is that message customisable from within the plugin? If not, I think it's somewhat unfortunate, but I wouldn't necessarily hold back on the screenshots because of it. A caveat might suffice?

rikvdkleij:
It looks like you compiled a project in another IntelliJ window.

epost:
Ok, I'll close this one then.

PR #26, opened 2015-08-28
Allow multiple developers

Oblosys:
Modifications to the project files, so they can be shared between plugin developers that have different minor IntelliJ SDK versions (without having dirty files in git that can be accidentally included in pull requests). It's not extremely elegant, but I've checked with JetBrains, and they confirmed there was no better way to do this.


rikvdkleij:
Because project setup is moved to sbt project this PR is not relevant anymore.

PR #36, opened 2016-04-23
omit code completion in string literal

### LINKED ISSUE #35: Auto completion issue
will show code completion hint dialog when typing inside a string literal.

report

zjhmale:
Fix #35 

@rikvdkleij 

btw, i think we should ignore `intellij-haskell/intellij-haskell.iml` cause contributors will use any version of IDEA. for me, i currently use IDEA 15.0.2 Community Edition, so i have to checkout this file before commit some changes.

but `intellij-haskell/intellij-haskell.iml` is really important in current plugin dev work flow, so maybe a `intellij-haskell/intellij-haskell.iml.example` should be tracked and ignore the real `.iml` file.

### DIFF
--- a/src/main/scala/intellij/haskell/code/HaskellCompletionContributor.scala
+++ b/src/main/scala/intellij/haskell/code/HaskellCompletionContributor.scala
@@ -54,7 +54,7 @@ class HaskellCompletionContributor extends CompletionContributor {
 
   private final val PlainPrefixElementTypes = Seq(HS_COMMENT, HS_NCOMMENT, HS_PRAGMA_START, HS_PRAGMA_END)
 
-  extend(CompletionType.BASIC, PlatformPatterns.psiElement(), new CompletionProvider[CompletionParameters] {
+  val provider = new CompletionProvider[CompletionParameters] {
     def addCompletions(parameters: CompletionParameters, context: ProcessingContext, originalResultSet: CompletionResultSet) {
       val project = parameters.getPosition.getProject
       val file = parameters.getOriginalFile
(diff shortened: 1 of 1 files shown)


rikvdkleij:
Yeah, the `iml` file is important for running/debugging the plugin inside IntelliJ. When I switched to `sbt` I did not find other solution.  I also use this (almost empty) plugin project to create deployment zip file.

Other solution is to write down in README how to create plugin project.

zjhmale:
i can not find a better way with sbt neither for now 😂 . but use gradle is more idiomic way, it will influence the dev work flow without too much configuration in IDEA, just in command lines. like https://github.com/intellij-rust/intellij-rust and https://github.com/zjhmale/intellij-clojure-pretty-symbol but switch to gradle seems a lot work to do, just some suggestions.

rikvdkleij:
Idea is change name of current iml file to `intellij-haskell.iml.2016.1` for example.

zjhmale:
👍  that was exactly what i mean, an example file should be tracked by git instread of the real `.iml` file
>An Idea is to change name of current iml file to intellij-haskell.iml.2016.1 for example.

zjhmale:
ready to be merged @rikvdkleij

rikvdkleij:
The proper way to solve this is to pass an ElementPattern implementation in calling extend() in HaskellCompletionContributor.

See for examples the IntelliJ Erlang plugin.

zjhmale:
done @rikvdkleij

rikvdkleij:
Code completion only for vars?

zjhmale:
sorry 😂 , but through the logs i just find `VARID_ID` token type, even tyep keywords, maybe i should add `CONID_ID`
>Code completion only for vars?

zjhmale:
sorry :joy:, add `CONID_ID` now, i can just find these two token types from logs.
>Code completion only for vars?

rikvdkleij:
Not only VARID_ID and CONID_ID but also VARSYM_ID and CONSYM_ID

zjhmale:
done @rikvdkleij .

and what kind of token is type of VARSYM_ID and CONSYM_ID, it's a little bit difficult to figure out through the bnf file 😂

rikvdkleij:
@zjhmale Thank you :+1:

PR #38, opened 2016-04-24
turn on hlint inspection by default

### LINKED ISSUE #37: Does hlint work?
Inspection by HLint;
Quick fixes for HLint suggestions;

How does it work? I don't see any suggestions.

zjhmale:
Fix #37 

@rikvdkleij 

and btw, maybe we should refactor https://github.com/rikvdkleij/intellij-haskell/blob/master/src/main/scala/intellij/haskell/inspection/HlintQuickfix.scala to get a better quick fix functionality.

for now it just doesn't work out sometimes.

report

### DIFF
--- a/intellij-haskell/META-INF/plugin.xml
+++ b/intellij-haskell/META-INF/plugin.xml
@@ -169,7 +169,7 @@
         <typedHandler implementation="intellij.haskell.code.HaskellTypedHandler" id="HaskellFile"/>
 
         <localInspection language="Haskell" shortName="HLintInspection" displayName="Inspection by HLint"
-                         groupName="Haskell" enabledByDefault="false" level="WARNING"
+                         groupName="Haskell" enabledByDefault="true" level="WARNING"
                          implementationClass="intellij.haskell.inspection.HlintInspectionTool"/>
 
         <lang.importOptimizer language="Haskell" implementationClass="intellij.haskell.code.HaskellImportOptimizer"/>
--- a/src/main/scala/intellij/haskell/external/Hlint.scala
+++ b/src/main/scala/intellij/haskell/external/Hlint.scala
@@ -31,7 +31,26 @@ object Hlint {
           p,
           Seq("--json", psiFile.getVirtualFile.getPath)
         )
-        deserializeHlintInfo(output.getStdout)
+        val infos = deserializeHlintInfo(output.getStdout)
+        infos.map((info: HlintInfo) => {
+          val found = info.from
+          val suggest = info.to.getOrElse("")
+          val hint = List(info.hint, "Found:", found, "Why not:", suggest).mkString("\n")
+          HlintInfo(
+            info.module,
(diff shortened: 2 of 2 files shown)


yury-pachin:
And it will be more useful with show what exactly found and suggest in popup.

zjhmale:
i'm working on this.
>And it will be more useful with show what exactly found and suggest in popup.

rikvdkleij:
Also replacing current code by the HLint suggested code is complex and needs a lot of testing with different scenarios.

zjhmale:
seems `ProblemDescriptorBase` does not support multiline? @rikvdkleij 

you mean what different scenarios?
>Also replacing current code by the HLint suggested code is complex and needs a lot of testing with different scenarios.

rikvdkleij:
I mean PSI tree of original code can be different than PSI tree of suggested code, as with code over multiple lines.

rikvdkleij:
Btw, how do you create those nice tiny screenshots?

zjhmale:
http://www.cockos.com/licecap/ a little screen capture tool here 😄 
>Btw, how do you create those nice tiny screenshots?

zjhmale:
can you show a test case here? i can not imagine the issue 😂 
>I mean PSI tree of original code can be different than PSI tree of suggested code, as with code over multiple lines.

rikvdkleij:
A solution could be to integrate `hindent` in plugin, so suggested code gets formatted well

zjhmale:
good point 👍  i will take a look at that later.
>A solution could be to integrate hindent in plugin, so suggested code gets formatted well

zjhmale:
ready to be merged @rikvdkleij , will improve quickfix use `hindent` next pr.

zjhmale:
poke @rikvdkleij

rikvdkleij:
@zjhmale 
Hmm, if hlint suggests to remove something the suggestion text is not proper. 

Did you see code in HlintQuickfix#getText ?

zjhmale:
Why not proper? the suggestion text just show what hlint inspect.
>Hmm, if hlint suggests to remove something the suggestion text is not proper.

nope , i will check this out tomorrow, so tired today 😂 
>Did you see code in HlintQuickfix#getText ?

rikvdkleij:
I mean, `to`can be empty. So in that case you display something as `Found [...] why not:`

zjhmale:
got it, i will fix that.
>I mean, tocan be empty. So in that case you display something as Found [...] why not:

rikvdkleij:
hlint is optional dependency so not by default enabled.

andrewufrank:
where can i switch off hlint? i did not see it mentioned anywhere in a setting.

by the way: hlint is not a perfect check for haskell code. it does not check the cabal file and thus complains for imports which occur with the same name in two packages (e.g. control monad which is in mtl and monads-tf). needs an annoying PackageImport clause.

Issue #55, opened 2016-11-08
Support for Stack hoogle


rikvdkleij:
Idea is to select expression in editor and call `stack hoogle -- [expression]`

zjhmale:
@rikvdkleij which way to you think is the best to kindly show the hoogle result to the users?

rikvdkleij:
Done. Also added navigation by Hoogle.

Issue #57, opened 2016-11-10
Add Intention actions


rikvdkleij:
For example to add type signature to where binding or to add parens around expression.

zjhmale:
@rikvdkleij

I just tried to  add type signature to where binding, but meet a few obstacles.

## What we already got

* Get type signature for a specific PsiElement, [findTypeInfoForElement](https://github.com/rikvdkleij/intellij-haskell/blob/master/src/main/scala/intellij/haskell/external/component/StackReplsComponentsManager.scala#L90).

* A pretty neat way to add type signatures for declarations, [TypeSignatureIntentionAction](https://github.com/rikvdkleij/intellij-haskell/blob/master/src/main/scala/intellij/haskell/annotator/HaskellAnnotator.scala#L247).

## What we are missing

* The element type of binding declaration itself and its parameters are all [HaskellTokenType.VAR_ID](https://github.com/rikvdkleij/intellij-haskell/blob/master/gen/intellij/haskell/psi/HaskellTypes.java#L181), so it's hard to determine whether current caret is focus on the binding declaration or just focus on a function parameter.

* Because the [where PsiElement](https://github.com/rikvdkleij/intellij-haskell/blob/master/gen/intellij/haskell/psi/HaskellTypes.java#L183) is isolated from its bindings, so it is hard to determine whether current caret is inside a where block or not.

* The binding and its type signature are both isolated in the Psi Structure, so it is also hard to check whether there is already a type signature for the where binding or not.

rikvdkleij:
Solution is to improve the psi tree by changing haskell.bnf.

zjhmale:
@rikvdkleij ok.

Issue #48, opened 2016-10-20
SDK creation fails if path ends in /stack


chris-martin:
When first setting this up, I tried to set `/home/chris/.nix-profile/bin/stack` as the path to Stack when creating the SDK, since that's my path to Stack. The first several times it failed silently; the dialog closed and no SDK was added. After a restart of IntelliJ, it started giving an error popup "The selected directory is not a valid home for Stack binary".

It took me a while to figure out that it's actually asking for the path of a directory that contains a stack executable, not the path of a stack executable. It probably ought to either be able to strip the `/stack` from the end on it or have better failure behavior.

intellij community 2016.2.4

intellij-haskell 1.0.0-beta

rikvdkleij:
Thanks for reporting. I will improve that in next version.

rikvdkleij:
I try to reproduce your problem. I do not find a way to select binary instead of folder.

There is already validation and most part is handled by IntelliJ code.

rikvdkleij:
I will change `Stack binary` to `Stack binary folder`

chris-martin:
I didn't use the file browser ui, I just pasted the path into the bar, and when I pressed enter, the dialog closed with no feedback.

rikvdkleij:
Okay, I could reproduce this by pasting path and directly after that, press enter. In that case file browser do not have time to go to location. No SDK is created in that case. 
The code which which is normally invoked for validation and so on, is not invoked. I do not know how to hook into there.

But when importing project by using the wizard, it will not allow to create project without SDK.

rikvdkleij:
Btw, same problem for other SDK's, for example Java SDK

rikvdkleij:
On second thought, I took another look to IntelliJ's code for creating SDK's and I have some idea how to change/override it and make it more solid. In thw new way Stack binary has to be selected. Makes it also better solution, for example, to support Windows.

But it will be breaking change....

rikvdkleij:
Fixed in beta 4.

chris-martin:
Just tried it in beta 5, makes a lot more sense to me now. Thanks!

Issue #12, opened 2015-01-21
Hole driven development support


rikvdkleij:

zjhmale:
Does the "Hole driven development" mean that auto generating expressions from type signatures?

rikvdkleij:
Yes, but also the other way around.
See http://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#typed-holes, http://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#partial-type-signatures and  http://matthew.brecknell.net/post/hole-driven-haskell/

zjhmale:
@rikvdkleij Thanks for the links, which comes to my mind is that we can only gain extra error messages using typed holes, but we already show the error messages right? so what exactly does this feature want to achieve?

rikvdkleij:
To apply suggested expression or type.

Btw, error can become warning by enabling: `-fdefer-typed-holes`

zjhmale:
Is there any ways to get suggested expression or type using typed holes? For the examples on the documents, they are just some messages and can not extract any useful infos to apply to the code on the editor.
>To apply suggested expression or type.

Thanks, I see the flag in the first link.
>Btw, error can become warning by enabling: -fdefer-typed-holes

rikvdkleij:
This blog contains some more examples:
http://lonelyproton.com/posts/36-getting-more-familiar-with-types-through-hole-driven-development-3

This issue is an idea........  :smile:

rikvdkleij:
@zjhmale 
>  and represent type infos for a specific typed hole, right?

Yes, you are right, that is doable

zjhmale:
@rikvdkleij this link is great, I think I have some plans about this feature in my mind now :smile: 
>This blog contains some more examples:
http://lonelyproton.com/posts/36-getting-more-familiar-with-types-through-hole-driven-development-3

rikvdkleij:
Current solution is good enough to support it

PR #59, opened 2016-11-19
Parens intention

zjhmale:
Integrate Add and Remove parens intentions. Related to #57 

screenshot


rikvdkleij:
Thank you very much for your PR. You have added an useful feature.

Btw, you added code to create element in HaskellPsiUtil. There is already similar code in HaskellElementFactory.

rikvdkleij:
About your new feature, I have added feature "Extend selection" (ctrl + w) to make it easier to select expression

zjhmale:
HaskellElementFactory is great, sorry I did not notice it before, I'll use it to create PsiElement next time :)
>Btw, you added code to create element in HaskellPsiUtil. There is already similar code in HaskellElementFactory.

Cool!
>About your new feature, I have added feature "Extend selection" (ctrl + w) to make it easier to select expression

rikvdkleij:
> HaskellElementFactory is great, sorry I did not notice it before, I'll use it to create PsiElement next time :)

Or maybe I will refactor it soon :smile:

zjhmale:
If you prefer, I'll send another PR to refactor this 😄 , I think you may need more time to fix #55, #48 and #12.
>Or maybe I will refactor it soon 😄

rikvdkleij:
#55 and #48 are done. #12 partly.

rikvdkleij:
@zjhmale 
I get NPE because `start` and `end` in `getSelectionStartEnd` can be null. You have to wrap them in Option.

zjhmale:
@rikvdkleij ok, I'll fix it right now.

rikvdkleij:
@zjhmale Adding parens does not work if expression is one character long. It turns around left and right parens.

zjhmale:
@rikvdkleij I will fix that.

zjhmale:
Hi @rikvdkleij, Add-Parens action for one character long expression works fine I think, what does `turns around` mean here?

screenshot

zjhmale:
@rikvdkleij I can reproduce this, I will fix that.

src/main/scala/intellij/haskell/psi/HaskellPsiUtil.scala
```diff
@@ -112,4 +117,23 @@ object HaskellPsiUtil {
...
+  }
+
+  def getRightParenElement(project: Project): PsiElement = {
+    createParensFromText(project, "add = (1 + 2)").filter(_.getNode.getElementType == HS_RIGHT_PAREN).head
+  }
+
+  def getSelectionStartEnd(psiElement: PsiElement, editor: Editor): (PsiElement, PsiElement) = {
+    val psiFile = psiElement.getContainingFile
+    val start = psiFile.findElementAt(editor.getSelectionModel.getSelectionStart)
+    val end = psiFile.findElementAt(editor.getSelectionModel.getSelectionEnd - 1)
```
rikvdkleij: Why `getSelectionEnd -1` and not just `getSelectionEnd`?

src/main/scala/intellij/haskell/intention/RemoveParensIntention.scala
```diff
@@ -0,0 +1,25 @@
...
+class RemoveParensIntention extends PsiElementBaseIntentionAction {
+  override def invoke(project: Project, editor: Editor, psiElement: PsiElement): Unit = {
+    val (start, end) = HaskellPsiUtil.getSelectionStartEnd(psiElement, editor)
+    start.delete()
+    end.delete()
+  }
+
+  override def isAvailable(project: Project, editor: Editor, psiElement: PsiElement): Boolean = {
+    val (start, end) = HaskellPsiUtil.getSelectionStartEnd(psiElement, editor)
+    psiElement.isWritable && start.getNode.getElementType == HS_LEFT_PAREN && end.getNode.getElementType == HS_RIGHT_PAREN
```
rikvdkleij: Why not available if begins or ends with paren? It's possible to have nested parens.

src/main/scala/intellij/haskell/psi/HaskellPsiUtil.scala
zjhmale: Thanks, `getSelectionEnd -1` should be wrong especially for one character.

src/main/scala/intellij/haskell/intention/RemoveParensIntention.scala
zjhmale: Because in `invoke` we just remove `start` and `end` elements, so this intention should only work when you select an expression which begins with left paren and ends with right paren.

rikvdkleij: Sorry, I mean the ones in AddParensIntention

zjhmale: Ah, Ok, that make sense, I will address this today 😄

PR #65, opened 2016-12-03
Show Stack SDK Version

zjhmale:
@rikvdkleij I think it is better to show Stack version explicitly in `project` view.

image

### DIFF
--- a/src/main/scala/intellij/haskell/sdk/HaskellSdkType.scala
+++ b/src/main/scala/intellij/haskell/sdk/HaskellSdkType.scala
@@ -33,7 +33,7 @@ class HaskellSdkType extends SdkType("Haskell Tool Stack SDK") {
 
   override def suggestHomePath(): String = "/usr/bin"
 
-  override def suggestSdkName(currentSdkName: String, sdkHome: String): String = "Haskell Tool Stack"
+  override def suggestSdkName(currentSdkName: String, sdkHome: String): String = "Haskell Tool Stack" + "-" + getVersionString(sdkHome)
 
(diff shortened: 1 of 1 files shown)


rikvdkleij:
@zjhmale Sorry, this does not work. When you upgrade Stack, it will still display the old version.

zjhmale:
@rikvdkleij Maybe the `suggestSdkName` method will be invoked only once when creating a new SDK? So it will not automatic synchonize the executable version.

rikvdkleij:
@zjhmale Yes, must be. I think it's easier/better to keep it as it was. Also because in Project settings the SDK is displayed with 2 version numbers.

rikvdkleij:
@zjhmale What do you think?

zjhmale:
@rikvdkleij Sorry for the delayed reply. After searching around, I found out that this approach already exists in other relatively sophisticated language plugins such as [go-plugin](https://github.com/go-lang-plugin-org/go-lang-idea-plugin/blob/master/src/com/goide/sdk/GoSdkType.java#L88) and [erlang-plugin](https://github.com/ignatov/intellij-erlang/blob/master/src/org/intellij/erlang/sdk/ErlangSdkType.java#L139). And the issue that displays with 2 version numbers in project settings seems inevitable as far as I know, the `go-lang-plugin` also has the same issue.

image

If you are still not happy with this change, maybe you can revert it, I'm ok with that :)

rikvdkleij:
> @rikvdkleij Sorry for the delayed reply.

No problem.

> If you are still not happy with this change, maybe you can revert it, I'm ok with that :)

I will revert it because it's confusing. I'm glad you are ok with the revert.

PR #69, opened 2016-12-08
Sync module name when rename file

zjhmale:
## Implemented

* Renaming a file should rename the module name without the module prefix.

screenshot

## Future work

1. Currently, I just ignore the references between `HS_MODID` and Haskell files [here](https://github.com/zjhmale/intellij-haskell/blob/bc9582666cccee8f9e538dd070cc99018fb36512/src/main/scala/intellij/haskell/psi/HaskellReferenceContributor.scala#L33) in order to make sure the renaming file action work as expected, but we can not rename module name anymore.

screenshot

  So maybe the references of `HS_MODID`  [here](https://github.com/rikvdkleij/intellij-haskell/blob/master/src/main/scala/intellij/haskell/navigation/HaskellReference.scala#L41) can be improved? In the old way, renaming a file and renaming module name will override the module prefix.

screenshot

2. Rename package name should also rename the module prefix, maybe [DirectoryAsPackageRenameHandlerBase](https://github.com/JetBrains/intellij-community/blob/master/platform/lang-impl/src/com/intellij/refactoring/rename/DirectoryAsPackageRenameHandlerBase.java) can do the job.

### DIFF
--- a/intellij-haskell/META-INF/plugin.xml
+++ b/intellij-haskell/META-INF/plugin.xml
@@ -181,6 +181,7 @@
                              implementationClass="intellij.haskell.refactor.HaskellNamesValidator"/>
         <lang.refactoringSupport language="Haskell"
                                  implementationClass="intellij.haskell.refactor.HaskellRefactoringSupportProvider"/>
+        <renamePsiElementProcessor implementation="intellij.haskell.refactor.HaskellRenameFileProcessor"/>
         <renamePsiElementProcessor implementation="intellij.haskell.refactor.HaskellRenameVariableProcessor"/>
         <vetoRenameCondition implementation="intellij.haskell.refactor.HaskellVetoRenameCondition"/>
 
--- a/src/main/scala/intellij/haskell/psi/HaskellReferenceContributor.scala
+++ b/src/main/scala/intellij/haskell/psi/HaskellReferenceContributor.scala
@@ -30,6 +30,7 @@ class HaskellReferenceContributor extends PsiReferenceContributor {
       @NotNull
       def getReferencesByElement(@NotNull element: PsiElement, @NotNull context: ProcessingContext): Array[PsiReference] = {
         element match {
+          case _: HaskellModid => PsiReference.EMPTY_ARRAY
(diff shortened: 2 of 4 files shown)


zjhmale:
@rikvdkleij Or maybe I can split this into two separated PR, so we can get `Adding module prefix when creating files in nested packages` merged first :)

rikvdkleij:
@zjhmale yes, good idea :+1:

rikvdkleij:
Very cool you acknowledged and solved issue with module prefix when creating Haskell file.

I also appreciate you extensive description with pictures.

I will take a look to the issue of `Future work` when I have time.

rikvdkleij:
Something else. Do you have experience with the Grammar Kit? I would like to improve the parser. For example to recognize where bindings (with type signatures) but also to find solution to recognize indentation. Currently, when a top level declaration and his binding are indented, the Haskell compiler will not complain but parser of plugin will fail. Same with expressions. Expressions have to be recognized by indentation and not by line endings.

zjhmale:
That will be great! Currently, I can not address issues in Future work by myself especially for Haskell Reference.
>I will take a look at the issue of Future work when I have time.

zjhmale:
@rikvdkleij I'm also trying to improve the parser to fix #57. Currently, I'm not an expert on Grammar-Kit and only have a little basic experience using it 😅. But I will definitely get familiar with it ASAP to assistant you after finishing the ongoing project I'm dealing with apparently.

>Something else. Do you have experience with the Grammar Kit? I would like to improve the parser. For example to recognize where bindings (with type signatures) but also to find a solution to recognize indentation. Currently, when a top level declaration and his binding are indented, the Haskell compiler will not complain but the parser of the plugin will fail. Same with expressions. Expressions have to be recognized by indentation and not by line endings.

rikvdkleij:
>@rikvdkleij I'm also trying to improve the parser to fix #57. Currently, I'm not an expert on Grammar-Kit >and only have a little basic experience using it :sweat_smile:. But I will definitely get familiar with it >ASAP to assistant you after finishing the ongoing project I'm dealing with apparently.

Thanks!

rikvdkleij:
Trying to reopen this PR.

PR #73, opened 2016-12-08
Sync module name when rename file

rikvdkleij:
Reverts rikvdkleij/intellij-haskell#72


rikvdkleij:
Originally https://github.com/rikvdkleij/intellij-haskell/pull/69

PR #70, opened 2016-12-08
Add module prefix when create new Haskell module

zjhmale:
Mentioned at #69.

## Implemented

* Adding module prefix when creating files in nested packages.

screenshot

PR #74, opened 2016-12-11
Refactor

zjhmale:
refactor for #70, @rikvdkleij please review.

### DIFF
--- a/src/main/scala/intellij/haskell/action/CreateHaskellFileAction.scala
+++ b/src/main/scala/intellij/haskell/action/CreateHaskellFileAction.scala
@@ -58,48 +58,54 @@ class CreateHaskellFileAction extends CreateFileFromTemplateAction(CreateHaskell
   }
 
   override def createFileFromTemplate(originName: String, template: FileTemplate, originDir: PsiDirectory): PsiFile = {
-    val pathItems = HaskellFileUtil.getPathFromSourceRoot(originDir.getProject, originDir.getVirtualFile).orNull
+    val pathItems = HaskellFileUtil.getPathFromSourceRoot(originDir.getProject, originDir.getVirtualFile)
 
-    // Adapted from super definition.
-    val mkdirs = new CreateFileAction.MkDirs(originName, originDir)
-    val name = mkdirs.newName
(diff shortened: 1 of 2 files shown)


src/main/scala/intellij/haskell/action/CreateHaskellFileAction.scala
```diff
@@ -58,48 +58,54 @@ class CreateHaskellFileAction extends CreateFileFromTemplateAction(CreateHaskell
...
-    val name = mkdirs.newName
-    val dir = mkdirs.directory
-    val project = dir.getProject
-
-    val nameWithmodulePrefix = if (pathItems == null || invalidPathItems(pathItems) || pathItems.isEmpty) {
-      name
-    } else {
-      pathItems.mkString(".") + "." + name
-    }
+    val pathItems = HaskellFileUtil.getPathFromSourceRoot(originDir.getProject, originDir.getVirtualFile)
```
zjhmale: @rikvdkleij can not use `originDir.getContainingFile` here, or there will be an NPE.

rikvdkleij: I see, weird, with files it's safer to use getContainingFile. The PsiDirectoryImpl just returns null...

rikvdkleij:
@zjhmale  I did small refactor of your code after merging.

zjhmale:
@rikvdkleij Great! I saw the commit [here](https://github.com/rikvdkleij/intellij-haskell/commit/eaf7a863d553451b1c3bf7d1ba543ad41ad671ba), and `foreach` in Scala is just the same as `fmap` in Haskell right?

rikvdkleij:
@zjhmale No, `foreach` is for side effects.

zjhmale:
@rikvdkleij ok, got it.

rikvdkleij:
@zjhmale Create Haskell file in test directory does not work anymore. It's related to this PR. I will fix it.

rikvdkleij:
I reverted all your code because I do not see easy fix and I have to continue with developing my Haskell application.

PR #75, opened 2016-12-11
Fix internalFileTemplate

zjhmale:
Related with #70.

@rikvdkleij Now the template in Settings/CodeTemplates is the same as the that is used when creating a new file.

image


rikvdkleij:
@zjhmale Thank you very much!

PR #84, opened 2016-12-20
Add Cabal file spellchecker

zjhmale:
screenshot

### DIFF
--- a/src/main/scala/intellij/haskell/spellchecker/CabalSpellcheckingStrategy.scala
+++ b/src/main/scala/intellij/haskell/spellchecker/CabalSpellcheckingStrategy.scala
@@ -1,0 +1,17 @@
+package intellij.haskell.spellchecker
+
+import com.intellij.psi.PsiElement
+import com.intellij.spellchecker.tokenizer.{SpellcheckingStrategy, Tokenizer}
+import intellij.haskell.cabal.CabalLanguage
+
+/**
+  * Provide spellchecker support for Cabal sources.
+  */
+class CabalSpellcheckingStrategy extends SpellcheckingStrategy {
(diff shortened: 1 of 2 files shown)


rikvdkleij:
Thanks!

src/main/scala/intellij/haskell/spellchecker/CabalSpellcheckingStrategy.scala
```diff
@@ -0,0 +1,17 @@
+package intellij.haskell.spellchecker
+
+import com.intellij.psi.PsiElement
+import com.intellij.spellchecker.tokenizer.{SpellcheckingStrategy, Tokenizer}
+import intellij.haskell.cabal.CabalLanguage
+
+/**
+  * Provide spellchecker support for Cabal sources.
+  */
+class CabalSpellcheckingStrategy extends SpellcheckingStrategy {
```
rikvdkleij: Can you explain me how this works? I do not see a dictionary file.

zjhmale: The document is [here](https://www.jetbrains.com/help/idea/2016.2/spellchecking.html), there is a built-in dictionary inside intellij, and [here](https://github.com/rikvdkleij/intellij-haskell/blob/master/src/main/resources/dictionary/haskell.dic) is a explcit dictionary for haskell words which we want to be accepted by the typo checker.

rikvdkleij:
That dict file for Haskell words is also used by the Cabal spellchecker?

zjhmale:
@rikvdkleij Yes, exactly, e.g. `aeson` will be a valid word in cabal file.
>That dict file for Haskell words is also used by the Cabal spellchecker?

PR #86, opened 2016-12-21
Improve cabal

zjhmale:
@rikvdkleij Address some issues of Cabal part you mentioned in the code review.

BTW, I still think [this](https://github.com/rikvdkleij/intellij-haskell/blob/master/src/main/scala/intellij/haskell/HaskellFile.scala#L34-L40) is a good approach, cause we need to use this functions couple times.


rikvdkleij:
Thanks!

rikvdkleij:
It would also be cool if we get auto completion for packages and it's version numbers. That is possible by using `cabal.config` of Stack LTS, for example: https://www.stackage.org/lts-7.14/cabal.config

zjhmale:
@rikvdkleij Cool, but is there any way to get the resolver from `stack.yaml`? Does intero has a command to do that?

rikvdkleij:
I do not know a direct way, only by extracting it from the path of `stack path --local-pkg-db`

About the version numbers. The version of package in LTS is fixed. So maybe we should only support autocompletion for package names.

zjhmale:
Thanks, I think it is a good approach.
>I do not know a direct way, only by extracting it from the path of stack path --local-pkg-db

Yes, or maybe we could complete the package name with version number, e.g. when the user type `ac`, then it will pop up following lookup elements.

```
ace ==0.6,
acid-state ==0.14.2,
action-permutations ==0.0.0.1,
active ==0.2.0.10,
```

>About the version numbers. The version of package in LTS is fixed. So maybe we should only support autocompletion for package names.

rikvdkleij:
> Yes, or maybe we could also complete the package name with version number, e.g. when the user type ac, then it will pop up following lookup elements.

No, I do not think it is good idea to mix those.

zjhmale:
@rikvdkleij I'll leave the version number and only popup package names then.

```
ace
acid-state
action-permutations
active
```

rikvdkleij:
Thanks!

PR #87, opened 2016-12-22
Improve redundant import intention

zjhmale:
@rikvdkleij Address issues of redundant import intention.

BTW, I still don't understand, why we only need ```[‘`](.+)[’']``` instead of ```[‘'`](.+)[’'`]``` in `DefinedButNotUsedPattern`. Do we miss a quote ```[']``` on the left and a quasi quote ```[`]``` on the right?


rikvdkleij:
Those different quotes are just from experiencing GHC messages on different OS and GHC versions. I did not see `'` yet :smiley:  We could add them to be safe...

zjhmale:
@rikvdkleij Yes, I will add them then.

PR #81, opened 2016-12-19
Support Cabal File

zjhmale:
The mature Cabal file parser is ported and adjusted from another great work [intellij-haskforce](https://github.com/carymrobbins/intellij-haskforce/tree/master/src/com/haskforce/cabal).

We all use `apache-2.0`, and there are no notices on the original parser files.

* Syntax highlighting and color setting page for Cabal file.
* Auto-completion for Haskell module.
* Haskell module navigation.

screenshot

PR #88, opened 2016-12-22
Resolve compiler warning

zjhmale:
Resolve eta-expansion of zero-argument method warnings which introduce since [scala2.12](http://www.scala-lang.org/news/2.12.0).

>Eta-expansion (conversion of a method to a function value) of zero-args methods has been deprecated, as this can lead to surprising behavior (#5327).

And currently I believe `() => foo()` is the safest way to invoke 0 argument method instead of `foo _`.

Related to #81.


rikvdkleij:
Thanks!

PR #89, opened 2016-12-22
Add more quotes

zjhmale:
Now the quotes group are all replaced by ```[‘’'`]```.

I think this should be seriously reviewed.

Discussed at #87.


rikvdkleij:
You should only add single quotes at start and end. As I wrote in #87

rikvdkleij:
On second thought, it is better to leave it as is because identifiers can also contain quotes. Hope you agree.

zjhmale:
@rikvdkleij Agreed, I'm also considering about the same thing, I will close this then, thanks!

PR #91, opened 2016-12-26
Cabal completion for package name

### LINKED ISSUE #90: Auto completion for package names
Discussed at #86.

zjhmale:
Fixes #90 

screenshot


src/main/scala/intellij/haskell/action/DownloadCabalConfigFile.scala
```diff
@@ -0,0 +1,17 @@
+package intellij.haskell.action
+
+import com.intellij.openapi.actionSystem.{AnAction, AnActionEvent}
+import intellij.haskell.external.component.HaskellComponentsManager.refreshGlobalProjectInfo
+import intellij.haskell.external.component.PathComponent
+
```
rikvdkleij: I do not see reason to create action for downloading cabal.config file. It should be downloaded when it is necessary.

Also no need to refresh.

I think best solution is to download file when project is created or opened.

zjhmale: Ok, so maybe download file [here](https://github.com/zjhmale/intellij-haskell/blob/cabal-completion/src/main/scala/intellij/haskell/external/component/GlobalProjectInfoComponent.scala#L58) in GlobalProjectInfo is cool enough? Cause the global project info will always be loaded when a project is created or opened right?
>I think best solution is to download file when project is created or opened.

I think the reason why we need this action is when the network is blocked maybe users can not download file at first, so when the network is available, users can use this action to download the config file manually and refresh the cache to make this completion working right.
>I do not see reason to create action for downloading cabal.config file. It should be downloaded when it is necessary.

zjhmale: Ok, so maybe download file [here](https://github.com/zjhmale/intellij-haskell/blob/cabal-completion/src/main/scala/intellij/haskell/external/component/GlobalProjectInfoComponent.scala#L58) in GlobalProjectInfo is cool enough? Cause the global project info will always be loaded when a project is created or opened right?
>I think best solution is to download file when project is created or opened.

I think the reason why we need this action is when the network is blocked maybe users can not download file at first, so when the network is available, users can use this action to download the config file manually and refresh the cache to make this completion working right.
>I do not see reason to create action for downloading cabal.config file. It should be downloaded when it is necessary.

rikvdkleij: >Ok, so maybe download file here in GlobalProjectInfo is cool enough? Cause the global project info will always be loaded when a project is created or opened right?

Yes, that's okay.

Well, then it's better to create or reuse existing action which refreshes everything. We could use the action which restarts the repls for this. We have many dependencies. Also user  can change dependending libraries or Stack LTS. We also have action for downloading sources. I will think about it...

zjhmale: Cool.
>Yes, that's okay.

Yes, currently the restart repl action will also trigger the config file check [here](https://github.com/rikvdkleij/intellij-haskell/blob/master/src/main/scala/intellij/haskell/action/RestartStackReplsAction.scala#L67), but restart repl will take a relatively long time, and the completion of package names did not reply on other things, so maybe extract the cabal file downloading action is more flexible for users to use?
>Well, then it's better to create or reuse existing action which refreshes everything. We could use the action which restarts the repls for this.

I think the change of Stack LTS is perceptible here cause this action will remove the old cabal config file (if it exists) and then download a new one.
>We have many dependencies. Also user can change dependending libraries or Stack LTS.

Right, we have `DownloadLibrarySources` already, so maybe another download action here is also reasonable?
>We also have action for downloading sources.

Take you time 😄 , if you still not happy with this action, I'll remove it then.
>I will think about it...

src/main/scala/intellij/haskell/external/component/PathComponent.scala
```diff
@@ -0,0 +1,80 @@
...
+      }.filter(!_.isEmpty).toList)
+    } catch {
+      case _: Exception => None
+    }
+  }
+
+  def downloadCabalConfig(project: Project): Option[Boolean] = {
+    getResolver(project).map(resolver => {
+      val url = s"https://www.stackage.org/$resolver/cabal.config"
+      val stdErr = CommandLine.runProgram(Some(project), getIdeaPath(project), "wget", Seq("--no-check-certificate", url), 10000, captureOutputToLog = true, logErrorAsInfo = true).map(_.getStderr)
```
rikvdkleij: You assume wget is on PATH.

src/main/scala/intellij/haskell/action/DownloadCabalConfigFile.scala
rikvdkleij: > Restarting the repls will also invaliYes, currently the restart repl action will also trigger the checking and downloading of cabal config file here, but restart repl will take a relatively long time, and the completion of package names did not reply on other things, so maybe extract the cabal file downloading action is more flexible for users to use?date GlobalCaches.

When package completion is requested and there is no cabal.config file it has to be downloaded. Because of the case user changes Stackage LTS release, every time project is opened the cabal.config has to be downloaded. Downloading library sources is only done at request and when creating project because it takes some time.

src/main/scala/intellij/haskell/external/component/PathComponent.scala
zjhmale: I will handle that.

src/main/scala/intellij/haskell/action/DownloadCabalConfigFile.scala
zjhmale: I think there is no need to download the cabal.config file every time a project is opened because we can get the old LTS release from the local cabal.config file (if it exists), so only when the cabal.config file is not downloaded or the resolver is changed, we have to download the file again. Downloading file is really time-consuming, so I believe this approach is reasonable enough. And another reason why this action exists here is that if users can not download the cabal.config file and load all the package names into global project info, they can just use this action directly without reopening the project, 
>When package completion is requested and there is no cabal.config file it has to be downloaded. Because of the case user changes Stackage LTS release, every time project is opened the cabal.config has to be downloaded. Downloading library sources is only done at request and when creating project because it takes some time.

rikvdkleij: I understand you. Problem is that user does not see relation between auto completion of package names and cabal.config file.

I think the most general solution is to download cabal.config file when creating project (just as library sources) and extend the action which restart the repls with invalidating/refreshing everything. So when user has problems because some plugin dependency is missing or for example he changes package dependencies in Cabal file, changes Stackage LTS version or changes language extensions; he can just use one action.

zjhmale: Ok, fair enough. I will remove the action then 😄. Users should not worry about some internal relations too much indeed.

zjhmale:
@rikvdkleij I just remove the redundant action and address the wget issue. Please review again :)

rikvdkleij:
I just scanned through your changes but I did not see how you address the wget issue.

zjhmale:
@rikvdkleij I think this [part](https://github.com/zjhmale/intellij-haskell/blob/3e3c56b824b3f54886f31eafc889bc9a1b7f84e1/src/main/scala/intellij/haskell/external/component/PathComponent.scala#L73-L85) is cool enough to handle wget stuff, if wget is not on PATH, it will throw a ProcessNotCreatedException and I found there is no stdout or stderr from wget so I use exit code to handle the process output.

rikvdkleij:
No, it is not good idea to just use wget and see if it works. You can use for example, `com.intellij.util.io.URLUtil#openStream`

zjhmale:
@rikvdkleij Good idea, it is always better `prevent user (as much as posible) from installing tools`, I will put this in my mind.
>No, it is not good idea to just use wget and see if it works. You can use for example, com.intellij.util.io.URLUtil#openStream

zjhmale:
@rikvdkleij Now I use `com.intellij.util.io.URLUtil#openStream` to replace the external tool, I think this PR is ready to be merged 😄

src/main/scala/intellij/haskell/external/component/PathComponent.scala
```diff
@@ -0,0 +1,119 @@
...
+object PathComponent {
+  private final val PackageNamePattern = """.* (.*) [==|installed].*""".r
+  private final val LocalPkgdbResolverPattern = """.*/(lts-\d+\.\d+)/.*\n""".r
+  private final val CabalConfigFileResolverPattern = """.*/(lts-\d+\.\d+)""".r
+
+  def getIdeaPath(project: Project): String = {
+    project.getBasePath + File.separator + ".idea"
+  }
+
+  def getConfigFilePath(project: Project): String = {
```
rikvdkleij: It's better to put cabal.config file in project's folder. As also mentioned in cabal.config file.

src/main/scala/intellij/haskell/external/component/PathComponent.scala
```diff
@@ -0,0 +1,119 @@
...
+
+  private def parseCabalConfigFile(project: Project): Option[Iterable[String]] = {
+    val configFilePath = getConfigFilePath(project)
+    try {
+      Some(Source.fromFile(configFilePath).getLines().filter(!_.startsWith("--")).map {
+        case PackageNamePattern(packageName) => packageName
+        case _ => ""
+      }.filter(!_.isEmpty).toList)
+    } catch {
+      case _: Exception => None
```
rikvdkleij: Write errror message to Event log.

src/main/scala/intellij/haskell/external/component/PathComponent.scala
```diff
@@ -0,0 +1,119 @@
...
+
+import com.intellij.openapi.project.Project
+import com.intellij.util.io.URLUtil
+import intellij.haskell.HaskellNotificationGroup
+import intellij.haskell.external.commandLine.StackCommandLine
+import org.apache.commons.io.FileUtils
+
+import scala.io.Source
+
+object PathComponent {
```
rikvdkleij: Why do you call it PathComponent?

src/main/scala/intellij/haskell/external/component/PathComponent.scala
```diff
@@ -0,0 +1,119 @@
...
+      Some(Source.fromFile(configFilePath).getLines().filter(!_.startsWith("--")).map {
+        case PackageNamePattern(packageName) => packageName
+        case _ => ""
+      }.filter(!_.isEmpty).toList)
+    } catch {
+      case _: Exception => None
+    }
+  }
+
+  def downloadCabalConfig(project: Project): Option[Boolean] = {
```
rikvdkleij: This method has only side effect, so result type is Unit

src/main/scala/intellij/haskell/external/component/PathComponent.scala
```diff
@@ -0,0 +1,119 @@
...
+      })
+    }
+  }
+
+  private def downloadAndParseCabalConfigFile(project: Project): Option[Iterable[String]] = {
+    downloadCabalConfig(project).flatMap(b => {
+      if (b) {
+        parseCabalConfigFile(project)
+      } else {
+        Some(List())
```
rikvdkleij: In case no result return None.

rikvdkleij:
https://github.com/zjhmale/intellij-haskell/blob/d8fabc9a1dc32f03d108b8e51e65df7f0f1516d6/src/main/scala/intellij/haskell/cabal/completion/CabalCompletionContributor.scala#L34 looks weird :smile: 

Why not call `completeExtensions` direct?
Why do you use qualified classes, like `psi.impl.ExtensionsImpl`?
You can also use filterNot two times.

src/main/scala/intellij/haskell/external/component/HaskellComponentsManager.scala
```diff
@@ -72,6 +72,10 @@ object HaskellComponentsManager {
     GlobalProjectInfoComponent.findGlobalProjectInfo(project)
   }
 
+  def refreshGlobalProjectInfo(project: Project): Unit = {
```
rikvdkleij: It's not used.

src/main/scala/intellij/haskell/external/component/PathComponent.scala
```diff
@@ -0,0 +1,119 @@
...
+  def getAllAvailablePackageNames(project: Project): Option[Iterable[String]] = {
+    val configFilePath = getConfigFilePath(project)
+    val configFile = new File(configFilePath)
+    if (!configFile.exists()) {
+      downloadAndParseCabalConfigFile(project)
+    } else {
+      val needUpdate = for {
+        oldResolver <- getResolverFromCabalConfigFile(project)
+        newResolver <- getResolverFromLocalPkgdb(project)
+      } yield oldResolver != newResolver
```
rikvdkleij: What is the scenario when oldResolver != newResolver?

src/main/scala/intellij/haskell/external/component/HaskellComponentsManager.scala
zjhmale: I'll remove that.

src/main/scala/intellij/haskell/external/component/PathComponent.scala
zjhmale: the old resolver is read from the existed cabal.config file and the new resolver is set by users, if users change the Stack resolver then `oldResolver != newResolver` will be true, just to make sure the cabal.config file can match up with the Stack resolver in the project.

src/main/scala/intellij/haskell/external/component/PathComponent.scala
zjhmale: Ok, fair enought.

src/main/scala/intellij/haskell/external/component/PathComponent.scala
zjhmale: Cool.

src/main/scala/intellij/haskell/external/component/PathComponent.scala
zjhmale: I'm also thinking about that if we could create a hidden folder called `.intellijhaskell`, then everything we downloaded through this plugin could be save there, e.g. the Haskell library sources.

rikvdkleij: Take also look to comments in the cabal.config itself  :smile:

zjhmale: Yes, I saw the comment there now.-- Please place this file next to your .cabal file as cabal.config
>As also mentioned in cabal.config file.

src/main/scala/intellij/haskell/external/component/PathComponent.scala
zjhmale: I think we can keep the return type here since we should use the return result to check whether we successfully download the cabal.config file.

rikvdkleij: Is not necessary for next step. You always have to check if file is there :smile:

zjhmale:
I'll fix these, thanks for reviewing 😄 

>https://github.com/zjhmale/intellij-haskell/blob/d8fabc9a1dc32f03d108b8e51e65df7f0f1516d6/src/main/scala/intellij/haskell/cabal/completion/CabalCompletionContributor.scala#L34 looks weird 😄

Why not call completeExtensions direct?
Why do you use qualified classes, like psi.impl.ExtensionsImpl?
You can also use filterNot two times.

zjhmale:
@rikvdkleij And I think the biggest reason I use qualified class [here](https://github.com/zjhmale/intellij-haskell/blob/d8fabc9a1dc32f03d108b8e51e65df7f0f1516d6/src/main/scala/intellij/haskell/cabal/completion/CabalCompletionContributor.scala#L41) because I can handle `Extensions`, `OtherExtensions` and `DefaultExtensions` [here](https://github.com/zjhmale/intellij-haskell/blob/d8fabc9a1dc32f03d108b8e51e65df7f0f1516d6/src/main/scala/intellij/haskell/cabal/lang/psi/CabalCompositeElement.scala#L76-L78) in a single case.

zjhmale:
@rikvdkleij Please review again.

rikvdkleij:
@zjhmale  Almost there :smile:

build.sbt
```diff
@@ -17,6 +18,7 @@ lazy val intellijHaskell = (project in file(".")).
     libraryDependencies += guava,
     libraryDependencies += scalaTest,
     libraryDependencies += sprayJson,
+    libraryDependencies += commonsIO,
```
zjhmale: Yes, I should not involve too many third dependencies 😅

zjhmale: @rikvdkleij BTW, where do you find those functions which JetBrains support internally, e.g `copyStreamToFile` and `openStream`, I think there is no Jetbrains official API documentation there 😅

rikvdkleij: Just using `Goto Symbol` and some guessing about the name :smiley:

zjhmale: :+1:

zjhmale:
@rikvdkleij I just port the `copyStreamToFile` to `HaskellFileUtil`, Please review again 😄

rikvdkleij:
Thanks!

PR #95, opened 2016-12-27
Use html highlight tag

zjhmale:
IntelliJ uses HTML syntax internally instead of markdown syntax, so maybe it is better to use HTML highlight tag in balloon notifications.

image

### DIFF
--- a/src/main/scala/intellij/haskell/external/commandLine/CommandLine.scala
+++ b/src/main/scala/intellij/haskell/external/commandLine/CommandLine.scala
@@ -48,10 +48,10 @@ object CommandLine {
 
     val processOutput = processHandler.runProcess(timeout.toInt, true)
     if (processOutput.isTimeout) {
-      HaskellNotificationGroup.logErrorBalloonEvent(project, s"Timeout while executing `${cmd.getCommandLineString}`")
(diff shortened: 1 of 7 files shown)


src/main/scala/intellij/haskell/external/commandLine/CommandLine.scala
```diff
@@ -48,10 +48,10 @@ object CommandLine {
 
     val processOutput = processHandler.runProcess(timeout.toInt, true)
     if (processOutput.isTimeout) {
-      HaskellNotificationGroup.logErrorBalloonEvent(project, s"Timeout while executing `${cmd.getCommandLineString}`")
+      HaskellNotificationGroup.logErrorBalloonEvent(project, s"Timeout while executing <b>${cmd.getCommandLineString}</b>")
       None
     } else if (!captureOutputToLog && !processOutput.getStderrLines.isEmpty) {
-      HaskellNotificationGroup.logErrorEvent(project, s"Error while executing` ${cmd.getCommandLineString}`:  ${processOutput.getStderr}")
+      HaskellNotificationGroup.logErrorEvent(project, s"Error while executing `${cmd.getCommandLineString}`:  ${processOutput.getStderr}")
       Option(processOutput)
```
rikvdkleij: I think you forgot this one

PR #96, opened 2016-12-27
Use exit code to check executables

zjhmale:
Currently if stack can not find an executable it will produce the stderr

>Executable named xxx not found on path: ...

So maybe it is better to check executables using exit code.

image

@rikvdkleij BTW, if I remove hlint from the path, and I can not locate the hlint executable using `which hlint` any more, `stack exec -- hlint` still works fine, can you explain the details under the hood?

### DIFF
--- a/src/main/scala/intellij/haskell/external/component/HLintComponent.scala
+++ b/src/main/scala/intellij/haskell/external/component/HLintComponent.scala
@@ -30,10 +30,8 @@ object HLintComponent {
   def check(psiFile: PsiFile): Seq[HLintInfo] = {
     val project = psiFile.getProject
     StackCommandLine.runCommand(Seq("exec", "--", HlintName, "--json", psiFile.getOriginalFile.getVirtualFile.getPath), project).map(output => {
-      if (output.getStderr.nonEmpty) {
-        if (output.getStderr.toLowerCase.contains("couldn't find file: hlint")) {
-          HaskellNotificationGroup.logWarningBalloonEvent(project, s"No Hlint suggestions because <b>$HlintName</b> build still has to be started or build is not finished yet.")
-        }
+      if (output.getExitCode == 1) {
+        HaskellNotificationGroup.logWarningBalloonEvent(project, s"No Hlint suggestions because <b>$HlintName</b> build still has to be started or build is not finished yet.")
       }
       deserializeHLintInfo(project, output.getStdout)
     }).getOrElse(Seq())
--- a/src/main/scala/intellij/haskell/external/component/HLintComponent.scala
+++ b/src/main/scala/intellij/haskell/external/component/HLintComponent.scala
@@ -31,7 +31,7 @@ object HLintComponent {
     val project = psiFile.getProject
     StackCommandLine.runCommand(Seq("exec", "--", HlintName, "--json", psiFile.getOriginalFile.getVirtualFile.getPath), project).map(output => {
       if (output.getExitCode != 0) {
-        HaskellNotificationGroup.logWarningBalloonEvent(project, s"Something went wrong while calling <b>$HlintName</b>. ${output.getStderr}. See event log for more info.")
+        HaskellNotificationGroup.logErrorBalloonEvent(project, s"Something went wrong while calling <b>$HlintName</b>. ${output.getStderr}. See event log for more info.")
       }
       deserializeHLintInfo(project, output.getStdout)
     }).getOrElse(Seq())
--- a/src/main/scala/intellij/haskell/external/component/HLintComponent.scala
+++ b/src/main/scala/intellij/haskell/external/component/HLintComponent.scala
@@ -30,8 +30,8 @@ object HLintComponent {
   def check(psiFile: PsiFile): Seq[HLintInfo] = {
     val project = psiFile.getProject
     StackCommandLine.runCommand(Seq("exec", "--", HlintName, "--json", psiFile.getOriginalFile.getVirtualFile.getPath), project).map(output => {
-      if (output.getExitCode == 1) {
-        HaskellNotificationGroup.logWarningBalloonEvent(project, s"No Hlint suggestions because <b>$HlintName</b> build still has to be started or build is not finished yet.")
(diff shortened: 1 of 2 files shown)


rikvdkleij:
> @rikvdkleij BTW, if I remove hlint from the path, and I can not locate the hlint executable using which hlint any more, stack exec -- hlint still works fine, can you explain the details under the hood?

Because I built hlint by Stack and not install it. Check `stack exec -- which hlint` in project's folder.

> So maybe it is better to check executables using exit code.

No, I do not think so.Writing the errors to console is already done in generic code. I just want to make distinction in errors between errors from tool or tool is not found. So this is special code to make plugin more user friendly.

zjhmale:
Cool, I can find the local executable now.
>Because I built hlint by Stack and not install it. Check stack exec -- which hlint in project's folder.

Ok, but I still can not understand in which scenario the process output will produce stderr `couldn't find file: xxx`
>No, I do not think so.Writing the errors to console is already done in generic code. I just want to make distinction in errors between errors from tool or tool is not found. So this is special code to make plugin more user friendly.

rikvdkleij:
For example when IntelliJ's inspection calls Hlint but Hlint build is not yet finished.

zjhmale:
Ok, so after the build process is done, stack still can not find hlint executable, then it will produce `Executable named hlint not found on path: ...` stderr right?
>For example when IntelliJ's inspection calls Hlint but Hlint build is not yet finished.

rikvdkleij:
Yes, or if build process is still busy or whatever reason the build is not completed.

zjhmale:
Ok, I'll check that and close this PR then :)
>Yes, or if build process is still busy or whatever reason the build is not completed.

zjhmale:
@rikvdkleij Please excuse me for asking that how can I reproduce the error message `couldn't find file: hlint`? I trigger hlint action while hlint is still under building but I can not get such an error, could you give me some hint? Thus I can kindly close this PR 😄

rikvdkleij:
What kind of error do you get when hlint is still bulding?

To reproduce that error you can remove hlint from path.

rikvdkleij:
Hmm, I now get different error message. It now contains:
`Executable named hlint not found on path`

zjhmale:
@rikvdkleij Yes, I got the same error message here, and that's why I initially opened this PR, you can revise my first comment and the screenshot. So the biggest reason why we need to check executables (e.g. hlint) using exit code is that the error message may still keep changing as the Stack itself moving on, but if the executable works fine then the exit code will always be 0, so maybe using exit code instead of matching error message strings is safer and reasonable 😄 
>Hmm, I now get different error message. It now contains:
Executable named hlint not found on path
Maybe it's message is changed in later version of Stack. We can "grep" on hlint not found?

rikvdkleij:
@zjhmale Okay, but then error balloon message should be more general, something like "Something went wrong while calling hlint. See event log for more info" :smile:

zjhmale:
I also embed the stderr in the balloon message, just show what is really going on directly for users 😄 
>@zjhmale Okay, but then error balloon message should be more general, something like "Something went wrong while calling hlint. See event log for more info" 😄

zjhmale:
@rikvdkleij One question beyond this PR, in the readme we say

>Plugin will automatically build Intero, Haskell-docs and HLint to prevent incompatibility issues. Those tools are built against Stackage release defined in project's stack.yaml; If you want to use later version of tool, you will have to build tool manually in project's folder by using stack build;

But actually, if I remove `hlint` and `haskell-docs excutables` from `stack exec` path, stack still can not find them and don't build them automatically, so what exactly does `Plugin will automatically build Intero, Haskell-docs and HLint` mean here? Thanks 😄

rikvdkleij:
Yes, my experience is that Stack will not do rebuild if you remove binaries. Removing the .stack-work directory triggers rebuild yes. I do not know other way.

src/main/scala/intellij/haskell/external/component/HLintComponent.scala
```diff
@@ -30,10 +30,8 @@ object HLintComponent {
   def check(psiFile: PsiFile): Seq[HLintInfo] = {
     val project = psiFile.getProject
     StackCommandLine.runCommand(Seq("exec", "--", HlintName, "--json", psiFile.getOriginalFile.getVirtualFile.getPath), project).map(output => {
-      if (output.getStderr.nonEmpty) {
-        if (output.getStderr.toLowerCase.contains("couldn't find file: hlint")) {
-          HaskellNotificationGroup.logWarningBalloonEvent(project, s"No Hlint suggestions because <b>$HlintName</b> build still has to be started or build is not finished yet.")
-        }
+      if (output.getExitCode != 0) {
+        HaskellNotificationGroup.logWarningBalloonEvent(project, s"Something went wrong while calling <b>$HlintName</b>. ${output.getStderr}. See event log for more info.")
```
rikvdkleij: Change to logError?

zjhmale:
Thanks 😄 
>Yes, my experience is that Stack will not do rebuild if you remove binaries. Removing the .stack-work directory triggers rebuild yes. I do not know other way.

src/main/scala/intellij/haskell/external/component/HLintComponent.scala
zjhmale: Sure, I change to `error` once, but then I rolled back for the code consistency 😅

rikvdkleij:
Thanks!

rikvdkleij:
I get this message as balloon and also in Event Log:
`Something went wrong while calling hlint. . See event log for more info.`
But the real cause is gone in Event Log :smile:

rikvdkleij:
Okay, Hlint gives info to stdOut, stdErr is empty but exit code is 1  :smile: 

I will change code.

zjhmale:
Ok, so hlint produce lint message with exit code 1, not quite intuitive 😅 
>@zjhmale I get this message as balloon and also in Event Log:
Something went wrong while calling hlint. . See event log for more info.
But the real cause is gone in Event Log 😄

PR #106, opened 2016-12-31
Specify source encoding in build.sbt

igrep:
This is required to build this project on my machine: Windows 10 in Japanese.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -22,3 +22,5 @@ lazy val intellijHaskell = (project in file(".")).
   )
 
 ideaBuild in ThisBuild := "163.7743.44"
+
+javacOptions ++= Seq("-encoding", "UTF-8")


build.sbt
```diff
@@ -21,4 +21,6 @@ lazy val intellijHaskell = (project in file(".")).
     unmanagedJars in Compile += baseDirectory.value / "idea"
   )
 
-ideaBuild in ThisBuild := "163.7743.44"
\ No newline at end of file
+ideaBuild in ThisBuild := "163.7743.44"
+
+javacOptions ++= Seq("-encoding", "UTF-8")
```
rikvdkleij: Why are you not adding them to javacOptions inside setting?

igrep: Yeah, it's already set in my IntelliJ.
Actually, I was stuck by the error caused by the source encoding when `compile` on sbt console.
Do you know how to set it on the console?
Sorry, I'm new to sbt.

rikvdkleij: My default encoding is UTF-8 in IntelliJ. I can not reproduce your issue. Does your change in build.sbt works for you?

rikvdkleij: Does it work for you if you add them to: `scalacOptions in Global`

rikvdkleij: @zjhmale Do you also have this encoding issue?

zjhmale: @rikvdkleij Nope, I currently work with OSX 10.10.5 and I got all my encoding in UTF-8.

image

zjhmale: @igrep Can you fix this with changing the default charset in Windows Terminal? IIRC the default encoding in Windows Console is not UTF-8.

igrep: Adding `%USERPROFILE%/.sbt/0.13/global.sbt` below worked! 😄 
Thank you!

```scala:global.sbt
javacOptions ++= Seq("-encoding", "UTF-8")
```

zjhmale: Great! :smile:

rikvdkleij: So we can close this PR?

zjhmale: LGTM
> So we can close this PR?

igrep: Yes. Thanks for the advice!

PR #114, opened 2017-01-05
Make ideaHaskellLib as hidden directory

### LINKED ISSUE #111: Hide `ideaHaskellLib` as `.ideaHaskellLib` or `.idea-haskell-libs`
Would be nice if the unpacked libraries stayed consistent with other reference/meta directories, hidden behind a `.`. Also think it would be better named as `.idea-haskell-libs`, but that's more of a personal preference.

zjhmale:
Fixes #111.


rikvdkleij:
Thanks!

PR #120, opened 2017-01-11
Download Nightly Cabal Config File

zjhmale:
Fix issue mentioned in gitter.

But @NightRa originally post this issue with

>CabalConfigComponent.getResolverFromLocalPkgdb, stack 1.3.2 returns "C:\Users\Ilan\Programming\Haskell-Playground.stack-work\install\ef7d292a\pkgdb", which contains no version.

The local pkgdb path neither contains resolver name or ghc version, in what scenario it will only contain a hash value?


rikvdkleij:
> The local pkgdb path neither contains resolver name or ghc version, in what scenario it will only contain a hash value?

I asked the same question on gitter.

rikvdkleij:
Thanks!

PR #126, opened 2017-01-11
Capitalize action name

zjhmale:
I think it is a convention that each word in action name should be capitalized 😄


rikvdkleij:
Thanks!

PR #124, opened 2017-01-11
Fix about page

zjhmale:
* Originally if I click the `About Haskell Project` button, it will introduce an NPE:

>null
java.lang.IllegalMonitorStateException
	at java.lang.Object.notify(Native Method)
	at intellij.haskell.HaskellNotificationGroup$.log(HaskellNotificationGroup.scala:104)
	at intellij.haskell.HaskellNotificationGroup$.balloonEvent(HaskellNotificationGroup.scala:98)
	at intellij.haskell.HaskellNotificationGroup$.logErrorBalloonEvent(HaskellNotificationGroup.scala:70)
	at intellij.haskell.settings.HaskellSettingsState$.notifyIfPathIsNotSet(HaskellSettingsState.scala:38)
	at intellij.haskell.settings.HaskellSettingsState$.getHindentPath(HaskellSettingsState.scala:26)
	at intellij.haskell.action.AboutAction.$anonfun$actionPerformed$1(AboutAction.scala:42)
	at intellij.haskell.action.AboutAction.$anonfun$actionPerformed$1$adapted(AboutAction.scala:35)
	at scala.Option.foreach(Option.scala:257)
	at intellij.haskell.action.AboutAction.actionPerformed(AboutAction.scala:35)
	at com.intellij.openapi.actionSystem.ex.ActionUtil$1.run(ActionUtil.java:199)
	at com.intellij.openapi.actionSystem.ex.ActionUtil.performActionDumbAware(ActionUtil.java:215)
	at com.intellij.openapi.actionSystem.impl.ActionMenuItem$ActionTransmitter.lambda$actionPerformed$2(ActionMenuItem.java:315)
	at com.intellij.openapi.wm.impl.FocusManagerImpl.runOnOwnContext(FocusManagerImpl.java:920)
	at com.intellij.openapi.wm.impl.IdeFocusManagerImpl.runOnOwnContext(IdeFocusManagerImpl.java:130)
	at com.intellij.openapi.actionSystem.impl.ActionMenuItem$ActionTransmitter.actionPerformed(ActionMenuItem.java:295)
	at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2022)
	at com.intellij.openapi.actionSystem.impl.ActionMenuItem.lambda$fireActionPerformed$0(ActionMenuItem.java:112)
	at com.intellij.openapi.application.TransactionGuardImpl.runSyncTransaction(TransactionGuardImpl.java:91)
[...]

PR #127, opened 2017-01-12
Highlight external tool name in about page

zjhmale:
Fixes issue mentioned at #124, Only bold tool names in OSX.


rikvdkleij:
Thanks!

PR #131, opened 2017-01-13
Get local resolver from stack.yaml file

### LINKED ISSUE #125: Support for hash value Stack resolver
This was originally proposed by @NightRa on Gitter.

>Played a little more with the plugin, it to parse output from stack on windows:
CabalConfigComponent.getResolverFromLocalPkgdb, stack 1.3.2 returns "C:\Users\Ilan\Programming\Haskell-Playground.stack-work\install\ef7d292a\pkgdb", which contains no version.
Also, assuming there must be an "lts-." is problematic - there are nightlies too

Currently this plugin support LTS and Nightly Stack resolver, but a hash value resolver may not relate to a specific cabal.config file.

zjhmale:
Fixes #125.

@rikvdkleij For a second thought, I believe there is no need to maintain a default cabal.config file in the plugin since each stack project should contain a validating resolver in its YAML file 😄


rikvdkleij:
> @rikvdkleij For a second thought, I believe there is no need to maintain a default cabal.config file in the plugin since each stack project should contain a validating resolver in its YAML file :smile:

I think that is not correct, my stack.yaml contains"
```
# Resolver to choose a 'specific' stackage snapshot or a compiler version.
# A snapshot resolver dictates the compiler version and the set of packages
# to be used for project dependencies. For example:
# 
# resolver: lts-3.5
# resolver: nightly-2015-09-21
# resolver: ghc-7.10.2
# resolver: ghcjs-0.1.0_ghc-7.10.2
# resolver:
#  name: custom-snapshot
#  location: "./custom-snapshot.yaml"
resolver: lts-7.14

```

zjhmale:
@rikvdkleij But the commented resolvers will not be retrieved, the on using resolver is lts-7.14.

rikvdkleij:
@zjhmale But it can be for example `ghc-7.10.2`, and that one does not refer to cabal.config.

zjhmale:
Ok, but everytime users reopen the project or restart stack repl, we will check whether the resolver from YAML file and cabal.config file are matching with each other.

rikvdkleij:
But if there is no cabal.config downloaded because stack.yaml contains from beginning `ghc` resolver.

zjhmale:
I got the point, sorry, I did not understand you at first 😅

zjhmale:
@rikvdkleij I think in other resolvers the stack repl will also encounter some problems, e.g. with ghc-7.10.2 resolver.

image

image

Have to use `stack solver` and then add those suggested deps to stack.yaml to make everything works fine.

zjhmale:
I think we should continue the original plan, after build intero manually, users can kindly use this plugin including package name completion with non lts or nightly resolvers. I just added notes on readme for reminding 😄 
>So no resolver but user still expects package name completion works 😄 Or we should not support that.

zjhmale:
@rikvdkleij Please review, a default cabal.config is added 😄

src/main/scala/intellij/haskell/external/component/CabalConfigComponent.scala
```diff
@@ -1,70 +1,88 @@
...
     downloadCabalConfig(project)
     parseCabalConfigFile(project)
   }
 
-  private def parseCabalConfigFile(project: Project): Option[Iterable[String]] = {
-    val configFilePath = getConfigFilePath(project)
+  private def parseCabalConfigFileBase(project: Project, lines: Iterator[String]): Option[Iterable[String]] = {
     try {
-      Some(Source.fromFile(configFilePath).getLines().filter(!_.startsWith("--")).map {
+      Some(lines.filter(!_.startsWith("--")).map {
```
rikvdkleij: Why wrap it in Some?

src/main/scala/intellij/haskell/external/component/CabalConfigComponent.scala
```diff
@@ -1,70 +1,88 @@
...
         case PackageNamePattern(packageName) => packageName
         case _ => ""
       }.filter(!_.isEmpty).toList)
     } catch {
       case _: FileNotFoundException =>
```
rikvdkleij: Where does the FileNotFoundException come from?

src/main/scala/intellij/haskell/external/component/CabalConfigComponent.scala
```diff
@@ -88,14 +105,18 @@ object CabalConfigComponent {
...
-        case _ => None
-      }
-    })
+  private def getResolverFromStackYamlFile(project: Project): Option[String] = {
+    try {
+      Option(new Yaml()
+        .load(new FileInputStream(new File(getYamlFilePath(project))))
+        .asInstanceOf[java.util.Map[String, Any]]
+        .get("resolver"))
+        .map(_.asInstanceOf[String])
```
rikvdkleij: What if value of "resolver" can not be found? get can return null.

src/main/scala/intellij/haskell/external/component/CabalConfigComponent.scala
```diff
@@ -1,70 +1,88 @@
...
 
-  private def parseCabalConfigFile(project: Project): Option[Iterable[String]] = {
-    val configFilePath = getConfigFilePath(project)
+  private def parseCabalConfigFileBase(project: Project, lines: Iterator[String]): Option[Iterable[String]] = {
     try {
-      Some(Source.fromFile(configFilePath).getLines().filter(!_.startsWith("--")).map {
+      Some(lines.filter(!_.startsWith("--")).map {
         case PackageNamePattern(packageName) => packageName
         case _ => ""
       }.filter(!_.isEmpty).toList)
```
rikvdkleij: Use filterNot

src/main/scala/intellij/haskell/external/component/CabalConfigComponent.scala
```diff
@@ -1,70 +1,88 @@
...
+            } else {
+              parseCabalConfigFile(project)
+            }
+          })
         }
-      })
-    }
+      } else {
+        val a = parseDefaultCabalConfigFile(project)
+        a
```
rikvdkleij: Refactor.

src/main/scala/intellij/haskell/external/component/CabalConfigComponent.scala
```diff
@@ -1,70 +1,88 @@
...
+        if (!configFile.exists()) {
           downloadAndParseCabalConfigFile(project)
         } else {
-          parseCabalConfigFile(project)
+          val needUpdate = for {
+            oldResolver <- getResolverFromCabalConfigFile(project)
+            newResolver <- getResolverFromStackYamlFile(project)
+          } yield oldResolver != newResolver
+
+          needUpdate.flatMap(b => {
```
rikvdkleij: What if needUpdate is None?

src/main/scala/intellij/haskell/external/component/CabalConfigComponent.scala
zjhmale: Sorry, just for debug purpose, forgot to remove that 😅

src/main/scala/intellij/haskell/external/component/CabalConfigComponent.scala
zjhmale: Yes, get can return null, that's why I wrap it up with a `Option`, I will also parse default cabal config file if can not get `resolver` from Stack YAML file 😄

zjhmale:
poke @rikvdkleij

src/main/scala/intellij/haskell/external/component/CabalConfigComponent.scala
rikvdkleij: But you are too late with wrapping in `Option` ....

rikvdkleij: O, wait, I do not saw second closing paren.

rikvdkleij:
Thanks!

PR #135, opened 2017-01-15
Add config file watcher to restart stack repls

### LINKED ISSUE #134: Config file listener
Show `Restart Stack REPLs` popup immediately after config files (stack.yaml, project.cabal, config.cabal) are changed.

Discussed at #130.

zjhmale:
Fixes #134.

screenshot

### DIFF
--- a/intellij-haskell/META-INF/plugin.xml
+++ b/intellij-haskell/META-INF/plugin.xml
@@ -279,6 +279,9 @@
                              implementationClass="intellij.haskell.refactor.HaskellNamesValidator"/>
         <renamePsiElementProcessor implementation="intellij.haskell.refactor.HaskellRenameVariableProcessor"/>
         <vetoRenameCondition implementation="intellij.haskell.refactor.HaskellVetoRenameCondition"/>
+
+        <!-- Notification -->
+        <editorNotificationProvider implementation="intellij.haskell.notification.ConfigFileWatcherNotificationProvider"/>
     </extensions>
 
     <application-components>
@@ -408,5 +411,4 @@
             <add-to-group group-id="HelpMenu" anchor="last"/>
         </action>
     </actions>
-
 </idea-plugin>
--- a/intellij-haskell/META-INF/plugin.xml
+++ b/intellij-haskell/META-INF/plugin.xml
@@ -296,7 +296,7 @@
             <implementation-class>intellij.haskell.external.repl.ProjectStackReplProcess</implementation-class>
         </component>
         <component>
-            <implementation-class>intellij.haskell.external.component.StackProjectStartupManager</implementation-class>
+            <implementation-class>intellij.haskell.external.component.StackProjectManager</implementation-class>
         </component>
 
         <component>
--- a/src/main/scala/intellij/haskell/action/AboutAction.scala
+++ b/src/main/scala/intellij/haskell/action/AboutAction.scala
@@ -30,7 +30,7 @@ class AboutAction extends AnAction {
 
   override def update(actionEvent: AnActionEvent) {
     val project = Option(actionEvent.getProject)
-    val isHaskellProject = project.exists(HaskellProjectUtil.isHaskellStackProject)
+    val isHaskellProject = project.exists(HaskellProjectUtil.isHaskellStackProject(_))
     actionEvent.getPresentation.setEnabled(isHaskellProject)
     actionEvent.getPresentation.setVisible(isHaskellProject)
   }
--- a/src/main/scala/intellij/haskell/action/RestartStackReplsAction.scala
+++ b/src/main/scala/intellij/haskell/action/RestartStackReplsAction.scala
@@ -21,8 +21,7 @@ import com.intellij.openapi.progress.Task.Backgroundable
 import com.intellij.openapi.progress.{ProgressIndicator, ProgressManager}
 import com.intellij.openapi.project.Project
 import intellij.haskell.HaskellNotificationGroup
-import intellij.haskell.external.commandLine.StackCommandLine
-import intellij.haskell.external.component.HaskellComponentsManager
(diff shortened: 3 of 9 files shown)


rikvdkleij:
When stack.yaml file is changed, restarting repls is not enough. The restart repls action is not same action as opening project. Opening project also (re)builds tool and rebuilds Hoogle database.

Take a look to logic in RestartStackReplsAction and StackProjectStartupManager.

rikvdkleij:
Also changing resolver in stack.yaml is much bigger change than for example, changing dependencies in cabal file.

I'm considering to only support changes in cabal file for restart repls. That means only rebuild hoogle database has to be added to RestartStackReplsAction.

rikvdkleij:
Btw, also notice that downloading library source files is only done when creating IntelliJ project.

zjhmale:
Yes, `moduleCreated` will only be called once when creating an IntelliJ project https://github.com/rikvdkleij/intellij-haskell/blob/master/src/main/scala/intellij/haskell/module/HaskellModuleBuilder.scala#L48. So maybe we should re-download library source files after cabal file is changed because users may update the dependencies.
>Btw, also notice that downloading library source files is only done when creating IntelliJ project.

zjhmale:
Do you mean changing resolver in stack.yaml will introduce some imcompatible issues?
>Also changing resolver in stack.yaml is much bigger change than for example, changing dependencies in cabal file.

rikvdkleij:
> Do you mean changing resolver in stack.yaml will introduce some imcompatible issues?

I mean, when resolver is changed, project and all tools have to be rebuild.

zjhmale:
@rikvdkleij So how about just stop REPLs and then call this part https://github.com/rikvdkleij/intellij-haskell/blob/master/src/main/scala/intellij/haskell/external/component/StackProjectStartupManager.scala#L39-L87 everytime we want to restart Stack REPLs when config files are updated or project SDK is changed?

rikvdkleij:
Okay, but I think it's not necessay when project SDK is changed.

zjhmale:
@rikvdkleij There is no need to rebuild tools when project SDK is changed?

rikvdkleij:
> @rikvdkleij There is no need to rebuild tools when project SDK is changed?

Yes, my first thought was not but maybe it's safe to do rebuild if also Project SDK is changed. It depends of course what is changed in Stack binary.

Okay, also do rebuild if Project SDK is changed.

zjhmale:
poke @rikvdkleij

zjhmale:
Two questions may be related to this PR, why we should sleep the thread manually for a second here https://github.com/rikvdkleij/intellij-haskell/blob/master/src/main/scala/intellij/haskell/action/RestartStackReplsAction.scala#L59, to wait for the cleanup process finish? And why we should start globalRepl again here https://github.com/rikvdkleij/intellij-haskell/blob/master/src/main/scala/intellij/haskell/action/RestartStackReplsAction.scala#L69, why restart globalRepl can release memory?

zjhmale:
Oh, I just got an error message like `timeout while executing stack clean`, is that because I did not sleep the thread manually like this https://github.com/rikvdkleij/intellij-haskell/blob/master/src/main/scala/intellij/haskell/action/RestartStackReplsAction.scala#L59?

rikvdkleij:
> Oh, I just got an error message like timeout while executing stack clean, is that because I did not sleep the thread manually like this https://github.com/rikvdkleij/intellij-haskell/blob/master/src/main/scala/intellij/haskell/action/RestartStackReplsAction.scala#L59?

No, it's because default timeout of StackCommandLine is too short. You can pass timeout to override default timeout.

rikvdkleij:
> Why we should sleep the thread manually for a second here https://github.com/rikvdkleij/intellij-haskell/blob/master/src/main/scala/intellij/haskell/action/RestartStackReplsAction.scala#L59, to wait for the cleanup process finish?

Just some pause to give repls time to stop.

rikvdkleij:
> Why we should start globalRepl again here https://github.com/rikvdkleij/intellij-haskell/blob/master/src/main/scala/intellij/haskell/action/RestartStackReplsAction.scala#L69, why restart globalRepl can release memory?

That is to release memory usage of global repl. When preloading cache with globel repl it takes a lot of memory.

zjhmale:
Thanks! I think I got the answers 😄

zjhmale:
@rikvdkleij So should I also add a manual pause after exit REPLs here https://github.com/zjhmale/intellij-haskell/blob/b38e47ad8cb0daac5231cd81514332f27a0c2abe/src/main/scala/intellij/haskell/external/component/StackProjectStartupManager.scala#L53?

rikvdkleij:
> @rikvdkleij So should I also add a manual pause after exit REPLs here https://github.com/zjhmale/intellij-haskell/blob/b38e47ad8cb0daac5231cd81514332f27a0c2abe/src/main/scala/intellij/haskell/external/component/StackProjectStartupManager.scala#L53 ?

Yes, but you can test if it works without manual pause. It's just some precaution.

zjhmale:
Ok, it all works fine without the manual pause at the moment, but it is safer to add it up I believe 😄 
>Yes, but you can test if it works without manual pause. It's just some precaution.

zjhmale:
poke @rikvdkleij

zjhmale:
poke @rikvdkleij

rikvdkleij:
After changing cabal file or Stack SDK the Stack repls restart does not check if another restart/build is already in process.

rikvdkleij:
And thanks for your PR! :smile:

rikvdkleij:
Also, it would be nice in both cases if you can ignore restart of repls.

rikvdkleij:
It looks like stack is changing the yaml file after build because I keep getting the `Config file is updated` notification even when I not change the file. I can not ignore it.....

PR #138, opened 2017-01-26
Add Libraries to External Libraries tree

### LINKED ISSUE #130: Add libraries to External Libraries tree
Just like Java, Kotlin, Scala, Go, Erlang plugin etc.

image

zjhmale:
Fixes #130

image

### DIFF
--- a/src/main/scala/intellij/haskell/module/HaskellModuleBuilder.scala
+++ b/src/main/scala/intellij/haskell/module/HaskellModuleBuilder.scala
@@ -21,11 +21,12 @@ import javax.swing.Icon
 
 import com.intellij.ide.util.projectWizard._
 import com.intellij.openapi.Disposable
+import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.module.{Module, ModuleType}
 import com.intellij.openapi.progress.{ProgressIndicator, ProgressManager, Task}
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.projectRoots.SdkTypeId
-import com.intellij.openapi.roots.{ModifiableRootModel, ModuleRootManager, ModuleRootModificationUtil, OrderRootType}
+import com.intellij.openapi.roots._
 import com.intellij.openapi.ui.Messages
 import com.intellij.openapi.util.Pair
 import com.intellij.openapi.util.io.FileUtil
@@ -174,7 +175,7 @@ class HaskellModuleWizardStep(wizardContext: WizardContext, haskellModuleBuilder
 
 object HaskellModuleBuilder {
 
-  private final val LibName = ".ideaHaskellLib"
+  private final val LibName = s".intellij-haskell${File.separator}lib"
   private final val DependencyPattern = """([\w\-]+)\s([\d\.]+)""".r
   private final val InitialProgressStep = 0.1
   private final val GhcPrimVersion = "0.5.0.0"
@@ -195,6 +196,7 @@ object HaskellModuleBuilder {
             val downloadedPackages = downloadHaskellPackageSources(project, stackPath, packages, progressIndicator)
             progressIndicator.setFraction(0.9)
             addPackagesAsLibrariesToModule(module, downloadedPackages, libDirectory.getAbsolutePath)
+            addPackagesAsLibrariesToSDK(module)
           })
         }
       })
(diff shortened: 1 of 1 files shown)


src/main/scala/intellij/haskell/module/HaskellModuleBuilder.scala
```diff
@@ -174,7 +175,7 @@ class HaskellModuleWizardStep(wizardContext: WizardContext, haskellModuleBuilder
 
 object HaskellModuleBuilder {
 
-  private final val LibName = ".ideaHaskellLib"
+  private final val LibName = s".intellij-haskell${File.separator}lib"
```
rikvdkleij: I do not understand this change. Why change name and use File.separator?

zjhmale: Because the name of `library root` can not be set manually, it will display the name of the last child directory, and I think `lib` is better than `.ideaHaskellLib` showing there, so I changed `LibName`, just for displaying `lib` in library tree under SDK.

image

rikvdkleij:
Thanks!

rikvdkleij:
@zjhmale Something goes terrible wrong:
 I had one Haskell project open, I created new project from existing sources, for that project it downloaded sources. In the other project External libraries also contains the packages of downloaded sources... The lib directory refers to directory of file system of already open project.

src/main/scala/intellij/haskell/module/HaskellModuleBuilder.scala
```diff
@@ -221,9 +223,12 @@ object HaskellModuleBuilder {
   private def downloadHaskellPackageSources(project: Project, stackPath: String, haskellPackages: Seq[HaskellPackageInfo], progressIndicator: ProgressIndicator) = {
     val step = 0.8 / haskellPackages.size
     var progressFraction = InitialProgressStep
-    haskellPackages.flatMap { packageInfo =>
-      val fullName = packageInfo.name + "-" + packageInfo.version
-      val stdErr = CommandLine.runProgram(Some(project), project.getBasePath + File.separator + LibName, stackPath, Seq("unpack", fullName), 10000, captureOutputToLog = true, logErrorAsInfo = true).map(_.getStderr)
+    haskellPackages.filter(packageInfo => {
```
rikvdkleij: I do not understand this change, you filter and filter not on same condition, so you end with same elements in haskellPackages..

rikvdkleij: I think your last two PR's are intermingled.

Also packages which are nog used anymore have to be deleted otherwise finding modules is not going work anymore because module names are not unique anymore.

zjhmale: That is because some part of the haskellPackages are no longer need to be downloaded but we still need to get all the haskell package names.
>I do not understand this change, you filter and filter not on same condition..

zjhmale: What do you mean by `finding modules is not going work anymore because module names are not unique anymore.`? can you give me more details? 😄 
>I think your last two PR's are intermingled.
Also packages which are nog used anymore have to be deleted otherwise finding modules is not going work anymore because module names are not unique anymore.

rikvdkleij: > That is because some part of the haskellPackages are no longer need to be downloaded but we still need to get all the haskell package names.

But the logic of managing source libraries is not right.

zjhmale:
@rikvdkleij I think reverting this PR is right since project SDK is shared with all projects, what I really want to implement is something like project dependencies, e.g. in intellij-scala, they do the trick here https://github.com/JetBrains/intellij-scala/blob/idea163.x/src/org/jetbrains/sbt/project/SbtProjectResolver.scala#L147

image

We need to implement an [ExternalSystemManager](http://www.jetbrains.org/intellij/sdk/docs/reference_guide/frameworks_and_external_apis/external_system_integration.html?search=External) instance to achieve that, and it is not wise to spend too much time just for implementing a minor feature, I will come back for this if we get all the key features done.

rikvdkleij:
@zjhmale it would be cool if you solve this issue because by this PR libraries are really added as libraries and not as project content

PR #139, opened 2017-01-27
Optimize downloading library sources

zjhmale:
Since #135 is taking redownloading library sources into consideration, so remove the library root directory and download all library sources is not a good idea anymore, it will cost user extra time to wait for the downloading process if there is a bunch of dependencies.

This PR handles two things

* Delete library sources which are no longer as dependencies of the project.
* Just download library sources which are not downloaded before.

### DIFF
--- a/src/main/scala/intellij/haskell/module/HaskellModuleBuilder.scala
+++ b/src/main/scala/intellij/haskell/module/HaskellModuleBuilder.scala
@@ -231,8 +231,8 @@ object HaskellModuleBuilder {
 
     haskellPackages.filter(packageInfo => {
       new File(project.getBasePath + File.separator + LibName + File.separator + packageInfo.dirName).exists()
-    }) ++ haskellPackages.filter(packageInfo => {
-      !new File(project.getBasePath + File.separator + LibName + File.separator + packageInfo.dirName).exists()
+    }) ++ haskellPackages.filterNot(packageInfo => {
+      new File(project.getBasePath + File.separator + LibName + File.separator + packageInfo.dirName).exists()
     }).flatMap { packageInfo =>
       val stdErr = CommandLine.runProgram(Some(project), project.getBasePath + File.separator + LibName, stackPath, Seq("unpack", packageInfo.dirName), 10000, captureOutputToLog = true, logErrorAsInfo = true).map(_.getStderr)
       progressFraction = progressFraction + step
--- a/src/main/scala/intellij/haskell/module/HaskellModuleBuilder.scala
+++ b/src/main/scala/intellij/haskell/module/HaskellModuleBuilder.scala
@@ -229,9 +229,10 @@ object HaskellModuleBuilder {
 
     libRoot.listFiles().filterNot(subDir => haskellPackages.map(_.dirName).contains(subDir.getName)).foreach(FileUtil.delete)
 
-    haskellPackages.filter(packageInfo => {
+    val alreadyDownloadedPackages = haskellPackages.filter(packageInfo => {
       new File(project.getBasePath + File.separator + LibName + File.separator + packageInfo.dirName).exists()
-    }) ++ haskellPackages.filterNot(packageInfo => {
+    })
+    val newDownloadedPackages = haskellPackages.filterNot(packageInfo => {
       new File(project.getBasePath + File.separator + LibName + File.separator + packageInfo.dirName).exists()
     }).flatMap { packageInfo =>
       val stdErr = CommandLine.runProgram(Some(project), project.getBasePath + File.separator + LibName, stackPath, Seq("unpack", packageInfo.dirName), 10000, captureOutputToLog = true, logErrorAsInfo = true).map(_.getStderr)
@@ -244,6 +245,7 @@ object HaskellModuleBuilder {
         Seq(packageInfo)
       }
     }
+    alreadyDownloadedPackages ++ newDownloadedPackages
   }
 
   private def getUrlByPath(path: String): String = {
--- a/src/main/scala/intellij/haskell/module/HaskellModuleBuilder.scala
+++ b/src/main/scala/intellij/haskell/module/HaskellModuleBuilder.scala
@@ -193,7 +193,7 @@ object HaskellModuleBuilder {
             val packageName = HaskellProjectUtil.findCabalPackageName(project)
             val packages = getPackages(project, dependencyLines.asScala).filterNot(p => packageName.contains(p.name))
             progressIndicator.setFraction(InitialProgressStep)
-            val downloadedPackages = downloadHaskellPackageSources(project, stackPath, packages, progressIndicator)
+            val downloadedPackages = downloadHaskellPackageSources(project, libDirectory, stackPath, packages, progressIndicator)
             progressIndicator.setFraction(0.9)
             addPackagesAsLibrariesToModule(module, downloadedPackages, libDirectory.getAbsolutePath)
           })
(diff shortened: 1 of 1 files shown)


rikvdkleij:
Cool but there are already conflicts and I do not see you also take version number of package into account.

zjhmale:
@rikvdkleij Confliction resolved, and I think the `dirName` of `HaskellPackageInfo` is already taking package version number into consideration here https://github.com/rikvdkleij/intellij-haskell/blob/master/src/main/scala/intellij/haskell/module/HaskellModuleBuilder.scala#L212

zjhmale:
poke @rikvdkleij, for now, it will not download multiple versions of one package.

zjhmale:
poke @rikvdkleij. It will remove all the downloaded packages which are not in the package list [here](https://github.com/zjhmale/intellij-haskell/blob/c03c43c7ff7e730cfc711a4afd54120c435837b9/src/main/scala/intellij/haskell/module/HaskellModuleBuilder.scala#L230), so I think there is no need to worry about holding multiple versions of one package in the `.ideaHaskellLib` library you mentioned yesterday 😄

zjhmale:
poke

src/main/scala/intellij/haskell/module/HaskellModuleBuilder.scala
```diff
@@ -218,12 +223,18 @@ object HaskellModuleBuilder {
...
+  private def downloadHaskellPackageSources(project: Project, libRoot: File, stackPath: String, haskellPackages: Seq[HaskellPackageInfo], progressIndicator: ProgressIndicator) = {
     val step = 0.8 / haskellPackages.size
     var progressFraction = InitialProgressStep
-    haskellPackages.flatMap { packageInfo =>
-      val fullName = packageInfo.name + "-" + packageInfo.version
-      val stdErr = CommandLine.runProgram(Some(project), project.getBasePath + File.separator + LibName, stackPath, Seq("unpack", fullName), 10000, captureOutputToLog = true, logErrorAsInfo = true).map(_.getStderr)
+
+    libRoot.listFiles().filterNot(subDir => haskellPackages.map(_.dirName).contains(subDir.getName)).foreach(FileUtil.delete)
+
+    haskellPackages.filter(packageInfo => {
```
rikvdkleij: I do not understand why you first filter on condition and then filter on not same condition.

rikvdkleij: I think you only need the filter not condition.

src/main/scala/intellij/haskell/module/HaskellModuleBuilder.scala
```diff
@@ -218,12 +223,18 @@ object HaskellModuleBuilder {
...
+      new File(project.getBasePath + File.separator + LibName + File.separator + packageInfo.dirName).exists()
```
rikvdkleij: It's safer and shorter to use: new File(project.getBasePath, libName).

rikvdkleij:
poke @zjhmale  :smile:

src/main/scala/intellij/haskell/module/HaskellModuleBuilder.scala
rikvdkleij: Sorry, but I still do not understand this logic.

You have to make distinction between the packages which have to be downloaded and which has to be added to library table. The library table is always cleared before adding library entries.

This is about downloading packages which are not downloaded yet.
```
 haskellPackages.filter(packageInfo => {
      new File(project.getBasePath + File.separator + LibName + File.separator + packageInfo.dirName).exists()
    }) ++ haskellPackages.filter(packageInfo => {
      !new File(project.getBasePath + File.separator + LibName + File.separator + packageInfo.dirName).exists()
    }).
```
is the same as
```
 haskellPackages.filterNot(packageInfo => {
      new File(project.getBasePath + File.separator + LibName + File.separator + packageInfo.dirName).exists()
    }).
```

zjhmale: @rikvdkleij These two code blocks behave differently. The first on will always return all Haskell packages and the second one will return part of the Haskell packages.

rikvdkleij: But in this way you will download packages which are already downloaded.

Like I said, 
> You have to make distinction between the packages which have to be downloaded and which has to be added to library table. The library table is always cleared before adding library entries.

zjhmale: Just as you said `The library table is always cleared before adding library entries.`, so if we only use `filterNot` part, the packages which are already downloaded will never be displayed in the library table.

rikvdkleij: > And then we just add these two parts and show all the packages in the library table, I still can not see any faults here :sweat_smile:

There are no two parts, there is a `++` between the two filtered lists.

Changing one to `filterNot` does not make a difference in issue.

zjhmale: Hi @rikvdkleij, just add some screenshot to make it more clear.

If we only use the filterNot part as you said

```
haskellPackages.filter(packageInfo => {
      !new File(project.getBasePath + File.separator + LibName + File.separator + packageInfo.dirName).exists()
    }).flatMap { packageInfo =>
      val stdErr = CommandLine.runProgram(Some(project), project.getBasePath + File.separator + LibName, stackPath, Seq("unpack", packageInfo.dirName), 10000, captureOutputToLog = true, logErrorAsInfo = true).map(_.getStderr)
      progressFraction = progressFraction + step
      progressIndicator.setFraction(progressFraction)

      if (stdErr.exists(_.contains("not found"))) {
        Seq()
      } else {
        Seq(packageInfo)
      }
    }
```

Those packages which are already downloaded will never be displayed on the module dependencies panel.

screen shot 2017-02-25 at 2 06 09 am

If we do what I just did in this commit, all the packages no matter it is already downloaded will be kindly displayed on the module dependencies panel.

screen shot 2017-02-25 at 2 08 57 am

rikvdkleij: I mentioned first only filterNot because I did not realised same list is used also for adding to library table.

rikvdkleij: You have to use filterNot to prevent packages are downloaded again.

rikvdkleij: I looked again to your code but you got it wrong. Otherwise I will pull your code and then change it in the right way 😄

rikvdkleij: Also, if you test it, you can look into the console to see that existing packages are downloaded again.

zjhmale: Yes, the same list will be displayed to library table.
>I mentioned first only filterNot because I did not realised same list is used also for adding to library table.

rikvdkleij: > I looked again to your code but you got it wrong. Otherwise I will pull your code and then change it in the right way :smile:

Otherwise this conversation never ends :smile:

zjhmale: Sorry, I don't get the point, you mean this part 

```Haskell
haskellPackages.filter(packageInfo => {
      new File(project.getBasePath + File.separator + LibName + File.separator + packageInfo.dirName).exists()
    })
```

should use `filterNot` instead?

>You have to use filterNot to prevent packages are downloaded again.

rikvdkleij: > Yes, the same list will be displayed to library table.

There have to be 2 lists, one for downloading packages which do not exist yet and one for adding to the library table.

zjhmale: Ok, give me sometimes, just let checkout the console 😄 
>Otherwise this conversation never ends 😄

zjhmale: Ok.
>There have to be 2 lists, one for downloading packages which do not exist yet and one for adding to the library table.

zjhmale: Hi @rikvdkleij, I have to say the current logic in my code is right, I can not see any part in the console complain about re-downloading packages which already downloaded. And the screenshot below can give the evidence that all the packages are downloaded in Fab 25th, and they will never be downloaded again only if I remove them.

screen shot 2017-02-26 at 2 41 27 pm
>Also, if you test it, you can look into the console to see that existing packages are downloaded again.

zjhmale: I think this is not a good approach since we always need to [check](https://github.com/rikvdkleij/intellij-haskell/blob/master/src/main/scala/intellij/haskell/module/HaskellModuleBuilder.scala#L235-L239) whether we can unpack those un-downloaded packages. But I just update the code and give each package list a name to make that code block more intelligible 😄 
>There have to be 2 lists, one for downloading packages which do not exist yet and one for adding to the library table.

zjhmale: How about putting your code snippet here? so we can discuss it in detail, I checked my code a lot of times and I'm 100% sure it is right 😅 
>I looked again to your code but you got it wrong. Otherwise I will pull your code and then change it in the right way 😄

zjhmale: We can end it today 😄 
>Otherwise this conversation never ends 😄

rikvdkleij:
Okay, now I see 😄  I saw it as one list. It was confusing, especially in github review page I do not see where expressions inside parens/brackets end.

Thanks!

zjhmale:
I will make my code more clean and intelligible next time, thanks for the reviewing Rik 😄 
>Okay, now I see 😄 I saw it as one list. It was confusing, especially in github review page I do not see where expressions inside parens/brackets end.

Thanks!

PR #145, opened 2017-02-06
Fix add paren intention for one character expression

zjhmale:
Fix issue mentioned at #59

screenshot

### DIFF
--- a/src/main/scala/intellij/haskell/intention/AddParensIntention.scala
+++ b/src/main/scala/intellij/haskell/intention/AddParensIntention.scala
@@ -3,7 +3,7 @@ package intellij.haskell.intention
 import com.intellij.codeInsight.intention.PsiElementBaseIntentionAction
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
-import com.intellij.psi.PsiElement
+import com.intellij.psi.{PsiElement, PsiWhiteSpace}
 import intellij.haskell.psi.HaskellTypes._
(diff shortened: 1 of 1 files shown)


rikvdkleij:
Thanks!

rikvdkleij:
@zjhmale It does not work if expression contains `.` an call paren intention action if cursor is on `.`

zjhmale:
Ok, I'll fix that!

zjhmale:
Hi Rik, I can not reproduce the issue on my side, can you give me some corner case code snippets to reproduce it?

sceenshot

>@zjhmale It does not work if expression contains . an call paren intention action if cursor is on .

zjhmale:
And you mean even the `add paren` intention action does not work?
>@zjhmale It does not work if expression contains . an call paren intention action if cursor is on .

rikvdkleij:
As your animated gif but without selecting the expression.

zjhmale:
Ok, I got it.

rikvdkleij:
Similar problem with for example `x:xs`

zjhmale:
Hi Rik, if the expected behavior is

`foo.bar` <=> `(foo.bar)`

`foo:bar` <=> `(foo:bar)`

when the cursor is on `.` and `:`.

I think it is quite impossible to implement at this moment because `foo.bar` is not the parent element of `.bar`. So the safest way to use these two intention actions is to apply them when we select the expression we want to apply to 😄

rikvdkleij:
Yes, that is expected behavior.

You can try to use nextSibling and/or prevSibling. 

But selecting is the expression is okay :smile:

PR #107, opened 2017-01-04
Support REPL

zjhmale:
Related to #99.

screenshot

Issue #11, opened 2015-01-21
Run (Hspec) tests inside IntelliJ


rikvdkleij:

rikvdkleij:
Fixed in beta11

Issue #129, opened 2017-01-12
Support Cabal flags per target/main


rikvdkleij:
Because project repl is loaded with `--test`, also option `--no-load` is passed to Stack repl otherwise we have to choose which target/main to load in Stack repl. 
Downside is that Cabal flags for target/main are not read, for example `ghc-options`

Possible solution is that each target (executable, test-suite) has it's own Stack repl.

rikvdkleij:
@zjhmale If we also create Stack intero repl for test target, we can run tests from there. Also no need to reload test hspec file when running test.

zjhmale:
@rikvdkleij Cool, but we should design an easy way for the users to communicate with multiple Stack REPLs right?

zjhmale:
And where can I find documentations about these options for `stack repl` command e.g. `--test` and `--no-load`? Thanks in advance 😄

zjhmale:
Oh, and one question beyond this issue, what is the use case of [GlobalStackReplProcess](https://github.com/rikvdkleij/intellij-haskell/blob/master/src/main/scala/intellij/haskell/external/repl/GlobalStackReplProcess.scala)? Because we will always open a Stack based project with this plugin, why we need a global one?

zjhmale:
@rikvdkleij https://gitlab.com/vannnns/haskero this VSCode extension support switching Intero process instance between targets defined in cabal file, I will do some investigation into this project in order to solve the issue here 😄

thoradam:
Haskero just has one intero process and restarts it if you select different targets. I thought about doing a process per target but I figured it required parsing the cabal file to map the source files to correct targets.

rikvdkleij:
Restarting repl takes some time. Consider switching between test and implementation files, I suspect some annoying delay Also the whole target has to be loaded again.

zjhmale:
Parsing cabal file is not pain in this plugin now, we can always extend the cabal parser if you need. And about the source file mapping process, I will investigate how `Haskero` do this.
>Haskero just has one intero process and restarts it if you select different targets. I thought about doing a process per target but I figured it required parsing the cabal file to map the source files to correct targets.

zjhmale:
Yeah, we should avoid restarting REPL too many times, let me do some investigation first 😄 
>Restarting repl takes some time. Consider switching between test and implementation files, I suspect some annoying delay Also the whole target has to be loaded again.

rikvdkleij:
Solved in beta 9

PR #150, opened 2017-02-21
Support REPL, Runner and Tester

### LINKED ISSUE #99: Support REPL and Runner
- [x] REPL
- [ ] Runner
- [ ] Run Test https://github.com/rikvdkleij/intellij-haskell/issues/11

zjhmale:
Fixes #99 and #11

This also fixes issues which mentioned at #107 

This PR basically address all the initial ideas

* Run a Stack project using `stack exec $executable` the `$executable` is configurable.
* Run a Stack project using `stack test $package:$testsuite` the `$testsuite` is configurable.
* Run Stack REPL inside a Stack project using `Stack ghci`.

The working directory and working module are also configurable.

Beyond the basic features, I also add a [RunStackCommandAction](https://github.com/zjhmale/intellij-haskell/blob/runconfig/src/main/scala/intellij/haskell/runconfig/action/RunStackCommandAction.scala) so we can easily just type `Ctrl/Cmd+Shilf+A` and the type `Run Haskell Stack command` and then simply enter the command we want to use, more details on the screenshot.

screenshot


rikvdkleij:
Cool!!

But before merging think it's better to first create repl per target. That would make plugin more solid because it prevent conflicts between targets (dependencies) and also tests can just be run in test repl.

zjhmale:
But fixing #129 may take longer time than we expected, so how about introducing these basic features first and when #129 is fixed we can improve this than 😃

src/main/scala/intellij/haskell/external/repl/ProjectStackReplProcess.scala
```diff
@@ -47,20 +48,24 @@ private[repl] class ProjectStackReplProcess(project: Project) extends StackReplP
   }
 
   private def load(psiFile: PsiFile, moduleName: Option[String]): Option[(StackReplOutput, Boolean)] = {
-    val filePath = getFilePath(psiFile)
-    val moduleNameAction = ApplicationManager.getApplication.runReadAction(new Computable[Option[String]] {
-      override def compute(): Option[String] = {
-        moduleName.orElse(HaskellPsiUtil.findModuleName(psiFile))
+    if (!HaskellConsoleView.isConsole(psiFile)) {
```
rikvdkleij: Can you explain this change?

zjhmale: Because [HaskellConsoleView](https://github.com/zjhmale/intellij-haskell/blob/4735b70def2a41f82f471eb760466aedd4a8bb1b/src/main/scala/intellij/haskell/runconfig/console/HaskellConsoleView.scala) will be treated as a PsiFile, so when you start an REPL, stack will load it as a common Haskell file just like the picture showing below

screen shot 2017-02-23 at 12 49 44 am

So we'd better skip this loading process for console view.

zjhmale: In #107 You also complained that

```
I also get this message in project repl:

12:04   [project-stack-repl] command: :load /λ Bla
12:04	[project-stack-repl] stdOut: ^IntellijHaskell^
11:50	[project-stack-repl] errOut: target ‘/λ’ is not a module name or a source file
```

Just the same reason here, so I add the `isConsole` predicate here 😄

rikvdkleij: Okay, but this is wrong place to check. Take a look to HaskellAnnotator.

Also, because it's a psi file, are you sure no other calls are made to this project  repl?

zjhmale: Thanks, I will.
>Okay, but this is wrong place to check. Take a look to HaskellAnnotator.

Yes, that is also the point that I was worry about, so is any general solutions that I can check this psi file in one place, I will check out `HaskellAnnotator` first 😄 
>Also, because it's a psi file, are you sure no other calls are made to this project repl?

zjhmale: @rikvdkleij Actually, the current approach is the best one I think, If I do the console file judgment here in [HaskellAnnotator](https://github.com/rikvdkleij/intellij-haskell/blob/master/src/main/scala/intellij/haskell/annotator/HaskellAnnotator.scala#L78), it will expose the psi file to `:load`, `:type-at`, `:loc-at`, but the changes here can avoid all these things.

image
image

rikvdkleij: You have to change `collectInformation` in `HaskellAnnotator`

Btw, I do not understand why the other commands are executed, looks like an accident.

zjhmale: Hi Rik, I just update `collectInformation` to 

```Haskell
  override def collectInformation(psiFile: PsiFile, editor: Editor, hasErrors: Boolean): PsiFile = {
    (psiFile, Option(psiFile.getOriginalFile.getVirtualFile)) match {
      case (pf, _) if HaskellConsoleView.isConsole(pf) => null
      case (_, None) => null // can be in case if file is in memory only (just created file)
      case (_, Some(f)) if f.getFileType != HaskellFileType.INSTANCE => null
      case (_, Some(_)) if HaskellProjectUtil.isLibraryFile(psiFile).getOrElse(true) => null
      case (_, Some(_)) => psiFile
    }
  }
```

But still can not prevent loading and checking console file, so my initial change seems still to be the best approach here.

rikvdkleij: I will take a look..

rikvdkleij: Your initial original solution does not handle the cause, it's a work around to prevent commands are executed on the project repl. We have to handle it earlier in the program flow.

zjhmale: Agreed, so where is the main entrance there?

rikvdkleij: Take a look where `HaskellProjectUtil.isLibraryFile()` is used.

zjhmale: Thanks, I will 😄

zjhmale: @rikvdkleij Should I check console file in each place where `HaskellProjectUtil.isLibraryFile()` is used? There are at lease 10 places where `HaskellProjectUtil.isLibraryFile()` stands 😅

rikvdkleij: Those are examples of entrances.  That is consequence of that Console is Haskell Psi file. On the other hand, you can use the code completion.

There is no higher entrance as far as I know. You can take a look in other plugins with repl, for example the Scala/sbt plugin.

src/main/scala/intellij/haskell/editor/HaskellCompletionContributor.scala
```diff
@@ -182,6 +184,10 @@ class HaskellCompletionContributor extends CompletionContributor {
                 resultSet.addAllElements(findTopLevelTypeSignatureLookupElements(psiFile).asJavaCollection)
               }
 
+              if (HaskellConsoleView.isConsole(psiFile)) {
```
rikvdkleij: We have to make separation in code between console file and normal file. Otherwise everything get mixed up and not maintainalbe anymore.

Also there is BrowseModuleComponent which has cached the top level identifiers for all modules.

zjhmale: Ok, maybe another completionContributor for `Haskell` language, is that possible to have multiple completionContributor for a single language?

src/main/scala/intellij/haskell/external/repl/ProjectStackReplProcess.scala
zjhmale: Yeah, I will get this done ASAP 😄

src/main/scala/intellij/haskell/editor/HaskellCompletionContributor.scala
rikvdkleij: The normal repl already has code completion. Did you check that in the Console?

rikvdkleij: > Ok, maybe another completionContributor for Haskell language, is that possible to have multiple completionContributor for a single language?

It's not necessary to have another completion contributor. From the entrance you have go to directly to another flow in which you probably can reuse already existing code and also the already cached exports. You will to set up in the same way, except that in the repl the loaded file is the context. Also depending with which target you start the repl.

zjhmale: The normal completion contributer can not work as expected, that' why I add that piece of code to add special result set for console file😄

rikvdkleij: Yes, I understand that, that is not my point of my latest 2 comments :smile:

rikvdkleij: In the console you do not know which Haskell file is loaded and which imports are manually added. That's what I mean by the context.

rikvdkleij: It would also be nice run just one Haskell test file (if it has a main)

rikvdkleij: > In the console you do not know which Haskell file is loaded and which imports are manually added. That's what I mean by the context.

If you add the action to load Haskell file in repl again, we know the context. Assuming the user does not load manually other file.

src/main/scala/intellij/haskell/runconfig/test/HaskellTestConfiguration.scala
```diff
@@ -45,8 +45,8 @@ class HaskellTestConfiguration(override val name: String, override val project:
   //https://github.com/commercialhaskell/stack/issues/2210
   override def getState(executor: Executor, environment: ExecutionEnvironment): HaskellStackStateBase = {
     val packageName: String = CabalQuery.getPackageName(project).getOrElse(project.getName)
-    val command = s"test $packageName:$myTestsuite" ++
-      (if (getTestFilter.isEmpty) "" else s" --test-arguments '-m${'"'}$getTestFilter${'"'}'")
-    new HaskellStackStateBase(this, environment, command)
+    val parameters = List("test", s"$packageName:$myTestsuite") ++
```
rikvdkleij: What do you mean by filter? I also not see the filter in the form when plugin is running

src/main/scala/intellij/haskell/editor/HaskellCompletionContributor.scala
```diff
@@ -392,6 +402,43 @@ class HaskellCompletionContributor extends CompletionContributor {
     waitForResult(lookupElements)
   }
 
+  private def getIdsFromAllExportedModules(project: Project) = {
```
rikvdkleij: You can use BrowseModuleComponent, also for external modules. Which modules to ask for, depends on the context.

zjhmale: OK, I will get it done tomorrow 😄

rikvdkleij: But how do you know context? I think the only useful solution is to make action to load current file in Console repl.

src/main/scala/intellij/haskell/runconfig/test/HaskellTestConfiguration.scala
zjhmale: Ah, the test filter is quite straightforward, it can fuzzy filter test cases by its description string, you can see the screenshot below, it is quite useful for production code base when it contains thousands of test cases, you won't want to run them all every time 😄 

screenshot

src/main/scala/intellij/haskell/editor/HaskellCompletionContributor.scala
zjhmale: OK, I will load current file as I did before, but what's the problem with `getIdsFromAllExportedModules`, because all exported identifiers can be accessed in any context if we start a stack repl process.

rikvdkleij: It only takes them for project files and we have ghci/intero command for it :smile:

src/main/scala/intellij/haskell/runconfig/test/HaskellTestConfiguration.scala
rikvdkleij: But in that case is much nicer to goto test file in IntelliJ and run test

src/main/scala/intellij/haskell/editor/HaskellCompletionContributor.scala
zjhmale: OK, so after loading current file in REPL, `BrowseModuleComponent` will kindly work as expected for me right? 😄

rikvdkleij: Yes, that's right :smile:

zjhmale: Actually, I should address the scenario that load another module after the repl is running, I will investigate BrowseModuleComponent soon, just a little busy these days😂

rikvdkleij:
This PR is merged after resolving conflicts. Because it's still open I will close it manually.

Issue #151, opened 2017-02-22
Line breaks in HLint intentions


zjhmale:
![screenshot](https://cloud.githubusercontent.com/assets/6234553/23222702/b761c7f2-f963-11e6-849b-e0c6d6f1932e.gif)

rikvdkleij:
Those line breaks are included in hlint suggestions. I will remove them. After applying you can use hindent to format code.

zjhmale:
Thanks Rik, I just planned to fix it this weekend 😄

rikvdkleij:
This solution does not work if original expression contains line breaks. For example in case expression.

zjhmale:
Yeah, it will replace all the `\n` with `\s` and the case expression wil be align into single line 😅

zjhmale:
But we do reformat it using hident after hlint intentions right?

rikvdkleij:
> But we do reformat it using hident after hlint intentions right?

Yes

zjhmale:
Wired, I will check this out later 😅

rikvdkleij:
> Yeah, it will replace all the \n with \s and the case expression wil be align into single line :sweat_smile:

This does not work if you multiple expressions within case.

zjhmale:
Yeah, right, it seems we got a dilemma here. Can you put some code snippet here, Rik?
>This does not work if you have multiple expressions within case.

rikvdkleij:
Redundant do:
```
foo x = do
  case x of
    Just y -> y
    Nothing -> 0
```

zjhmale:
One idea on my mind is that we only take care the case that the original code which need to be replaced is single line, and if the hlint intention is multiple line, then we need to replace `\n` as `\s+`, it just a trade off here 😅

rikvdkleij:
>  then we need to replace \n as \s+, it just a trade off here

Let's not replace line breaks if hlint intention is multiple line

zjhmale:
OK, I will do that 😄  just let hindent to handle the line break right?
>Let's not replace line breaks if hlint intention is multiple line

rikvdkleij:
Yes, let's give it a try, also with code of original issue :smile:

zjhmale:
Sure 😄

rikvdkleij:
Hmm. better solution is: only remove line breaks if original code is one line and intention is multiple

But I still do not see solid solution. Could you try solution which never removes line breaks and only apply hindent's formatting?

[...]

PR #155, opened 2017-02-24
Reformat using hindent after apply hlint quickfix

zjhmale:
Related to #151.

It is better to automatically trigger hindent formatting after applying hlint quick fix 😃

### DIFF
--- a/src/main/scala/intellij/haskell/action/HaskellFormatAction.scala
+++ b/src/main/scala/intellij/haskell/action/HaskellFormatAction.scala
@@ -31,7 +31,8 @@ class HaskellFormatAction extends AnAction {
       val selectionModel = actionContext.selectionModel
       selectionModel match {
         case Some(_) =>
-          HindentFormatAction.format(psiFile, selectionModel)
+          HindentFormatAction.format(psiFile, selectionModel.map(m =>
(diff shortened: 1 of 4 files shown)


rikvdkleij:
I do not think it is a good idea to always format the whole after applying suggestion. I was considering to use hindent only on the expression. Hindent can not always parse part of expresion

zjhmale:
Oh, yeah, good idea, I will do that 😄

zjhmale:
@rikvdkleij will only reformat with hindent for the commonParent expression now 😄

rikvdkleij:
Did you test this well? Because I have experience with `hindent` that it can only handle complete expressions. There is no guarantee that `commonParent`is a complete expression.

You can better use `HaskellPsiUtil.findExpressionParent`

zjhmale:
Thanks Rik, I noticed the problem and test this change in some cases, but yeah, using `findExpressionParent` is much more safer 😄 
>Did you test this well? Because I have experience with hindent that it can only handle complete expressions. There is no guarantee that commonParentis a complete expression.
You can better use HaskellPsiUtil.findExpressionParent

rikvdkleij:
Thanks!

PR #158, opened 2017-02-26
Eliminate hoogle error log

### LINKED ISSUE #157: Hoogle rebuild database failed
I have encountered this issue for a lot of times.

image

But something really wired here is that I can kindly rebuild hoogle database in command line interface with exactly the same project.

image

zjhmale:
Fixes #157.

Actually, maybe this is a hoogle upstream issue because it will treat these output below as stderr.

screen shot 2017-02-26 at 3 57 08 pm

So we should avoid it in order to make sure it will not introduce any confusions for users.

### DIFF
--- a/src/main/scala/intellij/haskell/external/component/HoogleComponent.scala
+++ b/src/main/scala/intellij/haskell/external/component/HoogleComponent.scala
@@ -72,7 +72,7 @@ object HoogleComponent {
     ApplicationManager.getApplication.executeOnPooledThread(new Runnable {
       override def run(): Unit = {
         try {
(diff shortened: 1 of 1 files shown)


rikvdkleij:
Thanks!

Btw, `CommandLine` has to be refactored because `logErrorAsInfo` only makes sense if `captureOutputToLog`  is `true`.

zjhmale:
Yeah, we can discuss it later 😄 
>Thanks!
Btw, CommandLine has to be refactored because logErrorAsInfo only makes sense if captureOutputToLog is true.

PR #159, opened 2017-02-27
Navigate to error location for `LoadProblemInOtherFile`

### LINKED ISSUE #152: Link to file for `LoadProblemInOtherFiles`
Currently, it is just some trivial error message there, maybe it is better to create a clickable link which references to the file contains that error since we can get the file path, the line number and the column number.

screen shot 2017-02-23 at 1 31 40 am

zjhmale:
Fixes #152.

screenshot


rikvdkleij:
Thanks!

PR #163, opened 2017-03-21
Parens intention improvement

zjhmale:
Address the issues discussed at #59.

### DIFF
--- a/src/main/scala/intellij/haskell/intention/AddParensIntention.scala
+++ b/src/main/scala/intellij/haskell/intention/AddParensIntention.scala
@@ -19,13 +19,7 @@ class AddParensIntention extends PsiElementBaseIntentionAction {
         end.getParent.addAfter(right, end)
       }
 
-      if (psiElement.getText.length == 1 && start.isInstanceOf[PsiWhiteSpace]) {
-        addParens(end, end)
-      } else if (psiElement.getText.length == 1 && end.isInstanceOf[PsiWhiteSpace]) {
-        addParens(start, start)
-      } else {
-        addParens(start, end)
-      }
+      addParens(start, end)
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/main/scala/intellij/haskell/intention/AddParensIntention.scala
```diff
@@ -19,21 +19,15 @@ class AddParensIntention extends PsiElementBaseIntentionAction {
...
+      addParens(start, end)
     }
   }
 
   override def isAvailable(project: Project, editor: Editor, psiElement: PsiElement): Boolean = {
     HaskellPsiUtil.getSelectionStartEnd(psiElement, editor) match {
       case None => false
       case Some((start, end)) =>
-        psiElement.isWritable && start.getNode.getElementType != HS_LEFT_PAREN && end.getNode.getElementType != HS_RIGHT_PAREN
+        psiElement.isWritable && (start.getNode.getElementType != HS_LEFT_PAREN || end.getNode.getElementType != HS_RIGHT_PAREN)
```
rikvdkleij: Why this condition: `(start.getNode.getElementType != HS_LEFT_PAREN || end.getNode.getElementType != HS_RIGHT_PAREN)`?

There is nothing wrong with adding extra parens and continue coding within parens, for example:
`(1 + 1)` then `((1 + 1))` and then `(1 + (1 + 1))` :smile:

zjhmale: OK, so we only check whether the psi element here is writable right?

rikvdkleij: Yes, btw how does this intention action works if you do not make selection?

zjhmale: It will wrap parenthesis around the identifier which the cursor is currently pointed to. I just tried last night.

rikvdkleij:
Thanks!

PR #184, opened 2017-05-06
Fix stylish-haskell only format action

vrom911:
Behavior of `Ctrl+Alt+U` hotkey was the same as the `Ctrl+Alt+;` - `hindent` was called anyway before `stylish-haskell`. It was due to bug (probably copy-paste error) in `plugin.xml`. This PR fixes behavior of `Ctrl+Alt+U` hotkey by calling appropriate format action.


rikvdkleij:
Thanks!

Issue #199, opened 2017-08-09
Parameter Info or persistent hints of some kind


solicode:
I'm finding that I want something when filling out long parameter lists/constructors. Right now I'm jumping between documentation/ghci/show type and code multiple times to do this. Is there currently a better way?

Ideally, I'd like IntelliJ's `Parameter Info` command to be implemented for Haskell, but I realize this would be quite difficult to do. I played with it a bit anyway using `ParameterInfoHandlerWithTabActionSupport` just to see what it would entail, and yes, it's definitely not trivial, especially with how Haskell syntax works (infix operators, pointfree notation, etc). A very hacky solution seems doable, but I don't know if there's value in a Parameter Info command that is wildly inaccurate in many situations...

So I gave up on that for now, and just tried the simple solution of having "Show Type Info" not auto-hide as you're typing. This addresses the original problem I was having with long parameter lists to some extent. I don't need to switch away from my main editor pane as much.

But I'd like to get your thoughts on this. Or if you have any other ideas that I'm not considering?

If you're okay with the idea though, I don't mind creating a PR for it. Though if I do that, I should probably make this configurable (or a separate IntelliJ command rather than modifying the default "Show Type").

rikvdkleij:
I understand your issue. Using the Type info is the best solution for now. Problem is that Type info is not present anymore when you are typing the function arguments. You have to give back to the function name and ask for the Type info.

I do not understand your solution completely. What do you mean by not-autohide. Do you want to show Type info in the status bar? But what if you are typing the function arguments?

Do you use the Type info shortcut?

solicode:
> Problem is that Type info is not present anymore when you are typing the function arguments. You have to give back to the function name and ask for the Type info.

Yes, this is what I'm referring to. My proposed change is that Type Info would no longer disappear while you're typing. The hint bubble stays on the screen. You have to hit ESC for the hint to disappear.

I have this working locally. But like I said, maybe it's not a good idea to make it the default behavior, so I would have to make it a config option or a separate command so that you can assign it to a different shortcut.

rikvdkleij:
Most simple solution is to create separate command/action.

Or implement it as parameter info. If the caret is not above function, find first previous sibling which is a function. Since beta9 also not compiling code should show Type info (but it is not 100% guaranteed for every function, for imported functions it should work) .

rikvdkleij:
[...]

PR #201, opened 2017-08-11
Add "sticky" type info action

solicode:
This is my attempt to add "sticky" Show Type Info as discussed in #199. Feel free to change the name. I didn't know what else to call it.

I tried to re-use already existing helper methods as much as I could, but I might have missed some. Particularly the whitespace handling was a little tricky because those nodes are on a different branch of the tree, so I had to "re-enter" the expression. I suspect there's a better way to do that, but I couldn't find anything to help me with that.

I'm sure this can be improved over time with better heuristics, but I kept the initial implementation simple for now.


rikvdkleij:
Thanks :+1:

PR #211, opened 2017-08-26
Extra REPL features

solicode:
I like the new REPL feature introduced in beta-11. I played a bit with it to see if I could make the REPL feedback/iteration loop a bit quicker with some additional features. This is mostly stuff I'm used to seeing in many tools for Lisp-like languages (one example being [Cursive](https://cursive-ide.com/userguide/repl.html)).

I don't know if you're interested or not. I mainly wanted to see what was possible. It's up to you whether any of this is something you're interested in including. I'm open to suggestions as well.

Anyway, to summarize the features:

- **File hyperlinks**: Clicking on errors in the console will take you to the line + column where it occurs
- **Load module in GHCi**: Loads the currently focused file in GHCi so you don't have to switch to the console and type ":load MyModuleName" and switch back.
- **Reload modules in GHCi**: Just a command to call ":reload" so you can bind it to a keyboard shortcut and not leave the editor pane.
- **Send to GHCi**: Sends the currently selected text to GHCi. If no text is selected, then it sends the current element at the cursor to GHCi. It also works with comments. You may want that for the case where you have a temporary scratchpad of commands that you want to run at a later time (since uncommenting them could cause your module to no longer compile).
- **Reload GHCi with last command**: Re-executes the last command after reloading your newest changes.

I didn't choose any keyboard shortcuts this time. I'm okay with anything there, or even just leaving them blank and letting the user choose.

Also just some miscellaneous thoughts:

- It would be nice if colored output for errors (introduced in GHC 8.2) were displayed rather than all the text being in red. I couldn't see a way of changing that though. I think IntelliJ always displays stderr in red regardless of any color codes?
- I'm not happy with the `lastCommand: Option[String]` thing I added, but I don't see a way of accessing the internal IntelliJ history model directly (barring reflection hacks).
- I went with the simplistic "execute the current element at the cursor" for the `Send to GHCi` command, but perhaps it could be made smarter by being context-sensitive. What I mean is, realizing the current element alone will fail in GHCi, so instead maybe selecting a large subexpression or something. I haven't given that much thought to be honest though. I just wanted to keep things as simple as possible for now.
- I noticed multiline commands aren't supported in the REPL. I didn't make any changes there, but it should be possible to handle that.
- stdout, stderr, stdin in the console window seem to interleave a bit sometimes. It's just a display thing, but it still bothers me. I'm not sure how to go about cleaning that up though.


solicode:
I got colorized error messages working locally. I can commit it later once it's fully complete. I would need to implement a version comparator (unless one already exists that I missed?) and branch if 8.2+ to apply the solution.

rikvdkleij:
Cool, thanks for your contribution!

> I would need to implement a version comparator (unless one already exists that I missed?) and branch if 8.2+ to apply the solution.

You can compare GHC version indirectly by reading Stack resolver:
https://github.com/rikvdkleij/intellij-haskell/blob/5d22e61c2dfa67eda8089c67ef2af188e9d8a0ff/src/main/scala/intellij/haskell/stackyaml/StackYamlComponent.scala#L14

solicode:
LTS doesn't yet have 8.2.1 (only nightly), so I would need to update that piece of the code in the future once it's added. Rather than do that, I ended up using `stack exec -- ghc --numeric-version`, which actually simplifies things in my opinion.

Also, redirecting the error stream seems to have fixed that interleaving issue I was talking about.

If there's anything else you'd like me to address in this PR, let me know.

I'll see if I can add the multi-line support eventually. For now, I'll create a separate issue for it.

rikvdkleij:
> Rather than do that, I ended up using stack exec -- ghc --numeric-version, which actually simplifies things in my opinion.

Yes, that can later in my mind. It's used in the Haskell About dialog.

rikvdkleij:
About `Send to GHCi`. What if REPL is not started or different module is loaded?


> I didn't choose any keyboard shortcuts this time. I'm okay with anything there, or even just leaving them blank and letting the user choose.

I will leave them blank because most short shortcuts are already occupied. User has to decide which to replace.

src/main/scala/intellij/haskell/external/execution/GhcVersion.scala
```diff
@@ -0,0 +1,14 @@
+package intellij.haskell.external.execution
```
rikvdkleij: Could be moved to HaskellProjectUtil together with the stack call to get GHC version. In this way it can be reused and I like to have the execution and parsing the result together.

solicode:
> What if REPL is not started

You're right, that case isn't handled. Right now if the REPL is not started, all of the GHCi commands do nothing, which probably isn't great UI... I think the simplest thing would be to disable the action if the project doesn't have a REPL started (`presentation.setEnabled(consoleView.isDefined)` in each `AnAction.update` method).

It's possible to do something fancier like try to start the REPL task automatically on first-run or something, but that requires more thought. For example, what if the user doesn't have a REPL task defined? Do you automatically create one? Or if they have multiple REPL tasks defined, you have to choose the right one.

> or different module is loaded?

Hmm, are you referring to a Haskell module or IntelliJ's concept of module? If it's the former, then I'm not sure I see a problem. If you try to run an expression in the wrong module, I'd expect the output to be `<interactive>:3:1: error: Variable not in scope: myFunction` or whatever the error is. The user would have to load the right module before running that expression. That's one reason I added "Load module in GHCi". Whenever I work in a different module, I use that command before doing work.

Though GHCi does have a concept of a "module set". Not sure if that changes anything. To be honest, I'm not familiar with it. But if it's important, maybe a "Add to module set" action could also be added? I might be misunderstanding how it works though.

> Could be moved to HaskellProjectUtil

Sure, I'll move that.

solicode:
There is another weird case I noticed. It's possible to start multiple REPL windows for the same project. I'm not sure why somebody would do that though. It's a user-defined task, so I don't see a way of forcing the "Single instance only" checkbox.

rikvdkleij:
>  I think the simplest thing would be to disable the action if the project doesn't have a REPL started

I agree, keep it simple.

> Hmm, are you referring to a Haskell module or IntelliJ's concept of module? 

I was referring to Haskell module. I can live with your explanation :smile: It is also no solution to keep module in editor and module in REPL in sync. Maybe the user does not even want that. It's nice that your contribution makes it very easy to load module in REPL.

rikvdkleij:
> There is another weird case I noticed. It's possible to start multiple REPL windows for the same project. I'm not sure why somebody would do that though. It's a user-defined task, so I don't see a way of forcing the "Single instance only" checkbox.

I think it's impossible to prevent that user starts multiple REPLs. It's even eligible if user want to starts REPL for library code and one for test code. A REPL is started for a specific stack component.

I assume that if user sends code the REPL, it will be sent to the active REPL.

solicode:
Ok, I made those changes.

> I assume that if user sends code the REPL, it will be sent to the active REPL.

It wasn't before, but I updated that too.

rikvdkleij:
Thanks!

PR #216, opened 2017-09-01
Clean Hoogle output for documentation

solicode:
Hoogle outputs excessive newlines in many cases (3 or more). For example, if you run `hoogle -i catMaybes` you should see what I'm referring to.

I think this occurs because Hoogle cleans up html tags, but doesn't touch up all the newlines that this leaves behind?

Anyway, here is a quick fix for that.

Also, `getStdout` is NotNull, so I removed the unnecessary null check.


rikvdkleij:
Thanks!

PR #223, opened 2017-10-02
Add multi-line REPL support

### LINKED ISSUE #212: Multi-line support in REPL
Currently, anything sent to the REPL is processed line-by-line, so sending a single expression that spans multiple lines doesn't work.

Detecting multi-line commands and automatically wrapping it in a `:{` `:}` block is one solution to this.

The only tricky part might be that GHCi also has a `+m` option for allowing multi-line commands. If the user has that option set (perhaps by a custom `.ghci`), the behavior is different. So maybe it's a good idea to consider that case too and make sure it works regardless of mode.

solicode:
Fixes #212 

Some notes about the implementation so far:

- `prompt2` looks like it was renamed to `prompt-cont` in GHC 8.2.1. To get rid of that `Some flags have not been recognized: prompt2,` warning, I guess I would need to create 2 separate GHCi config files.
- Rather than use a GHCi config file, it would be nice if I could set them with plain CLI flags (or environment variables)... but I couldn't find any such options in GHCi. It looks like the only choices are to use a config file, or send each command one by one after the REPL starts (like ":set +m", ":set prompt ...", and so on).

After some thinking, I'm actually thinking the latter is better, because that way I only override the relevant options and don't mess with any other custom settings a user has in their `.ghci` file. Because I really don't want to override *all* user ghci settings. I only want to change the problematic ones for IntelliJ.

### DIFF
--- a/src/main/resources/ghci/8.2.1.ghci
+++ b/src/main/resources/ghci/8.2.1.ghci
@@ -1,3 +1,3 @@
 :set +m
 :set prompt ""
-:set prompt2 ""
+:set prompt-cont ""
--- a/src/main/resources/ghci/default.ghci
+++ b/src/main/resources/ghci/default.ghci
@@ -1,0 +1,6 @@
+:set +m
+:set prompt ""
+:set prompt2 ""
+
+-- GHC 8.2.1+
+:set prompt-cont ""
--- a/src/main/scala/intellij/haskell/runconfig/console/HaskellConsoleState.scala
+++ b/src/main/scala/intellij/haskell/runconfig/console/HaskellConsoleState.scala
@@ -24,9 +24,15 @@ class HaskellConsoleState(val configuration: HaskellConsoleConfiguration, val en
     HaskellSdkType.getStackPath(project) match {
       case Some(stackPath) =>
         val stackTarget = configuration.getStackTarget
-        val ghciOptionsPath = getClass.getResource("/ghci/default.ghci").getPath
+        val ghcVersion = HaskellProjectUtil.getGhcVersion(project)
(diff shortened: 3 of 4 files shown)


rikvdkleij:
Thanks!

Yes, I also prefer second solution.

Btw, I want to release new version of plugin this week. Shall I wait merging this PR till it's completely finished?

solicode:
I don't think I'll be able to get to it today. But if it's just that one change I mentioned (the second solution), then I think I should be able to finish it by tomorrow.

rikvdkleij:
Tomorrow is fine. Thanks!

solicode:
Setting the options automatically through the console didn't work out properly. There were too many timing issues to worry about. If I send it at startup, that's too early and the commands get ignored. If I wait and set it lazily on the first user command, then the settings don't apply until the 2nd command. Very frustrating. And I couldn't find any lifecycle events to hook into or anything like that. And monitoring the output stream or something like that just seems like overkill for this.

Instead, I went with the ghci file option. Actually, it turns out that you can specify multiple ghci files (I didn't realize that was possible) and they layer just fine. This works because `~/.ghci` loads before `-ghci-script`. In fact, Stack seems to use this too (at least I think that's what the `Loaded GHCi configuration from /private/var/folders/jd/n_pz6bks6r3ffjcw4t53fl9w0000gn/T/ghci8162/ghci-script` is) [here](https://github.com/commercialhaskell/stack/blob/ac8e40c781e7b6e01bdec613c7a8ed1daef1b185/src/Stack/Ghci.hs#L362-L363).

So all I really needed to do was remove the `-ignore-dot-ghci` that I had added.

I also fixed the GHC 8.2.1 prompt2 rename warning.

rikvdkleij:
Thanks!

PR #226, opened 2017-10-05
Add documentation on how hindent is used

garetht:
This PR adds several sentences noting how `hindent` takes its options from internal IntelliJ settings.


rikvdkleij:
Thanks!

PR #230, opened 2017-10-11
Use temporary file for ghci-script

### LINKED ISSUE #229: Startup GHCi script not loading
This one is my fault. I forgot to take into account that the GHCi script would be inside the jar file. In dev mode it must have worked because it wasn't yet packaged up in a jar.

The problem line is [here](https://github.com/rikvdkleij/intellij-haskell/blob/5599bbb1e35e4184d6a4b89bfd57e51f1ba8de55/src/main/scala/intellij/haskell/runconfig/console/HaskellConsoleState.scala#L30-L32).

I guess the solution is to create a temp file like Stack does? Unless there's a different IntelliJ way of handling external files like this that I'm not aware of.

solicode:
Fixes #229

I kept this minimal, but I suppose it could be made smarter as to not create a temp file each time you start a new REPL (perhaps caching + using tempDirectory with fixed ghci files).

### DIFF
--- a/src/main/scala/intellij/haskell/runconfig/console/HaskellConsoleState.scala
+++ b/src/main/scala/intellij/haskell/runconfig/console/HaskellConsoleState.scala
@@ -6,8 +6,9 @@ import com.intellij.execution.filters.TextConsoleBuilderImpl
 import com.intellij.execution.process.{ProcessHandler, ProcessTerminatedListener}
 import com.intellij.execution.runners.ExecutionEnvironment
 import com.intellij.execution.ui.ConsoleView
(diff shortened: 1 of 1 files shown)


rikvdkleij:
Thanks for solving this issue so quickly! I will release new version asap. First I want to solve another issue.

rikvdkleij:
Suggestion: I have created an ghci options file for the background REPLs and have put it in the `~/.intellij-haskell` folder so it can be reused. Maybe also an idea for your REPL code.

rikvdkleij:
Thanks! No problem, I have not done major changes last couple of months 😄

PR #253, opened 2018-01-18
Fix execution bug in REPL panel

### LINKED ISSUE #239: GHCi commands sometimes don't execute
I still haven't properly diagnosed the cause of this, but every once in a while the GHCi actions (like reload modules) don't execute. I mean, the `:reload` shows up in the REPL panel, but it doesn't run. At least not until I run another REPL command. After that, it seems to work again.

I suspect it might be the `ConsoleViewContentType.USER_INPUT` change in PR #223. Maybe reverting back to `processInputWriter` is necessary. It's like the `\n` isn't interpreted right away in `ConsoleViewContentType.USER_INPUT` mode. That's just a guess though. I can look into it.

solicode:
Fixes #239

I built the fix on top of the old code because `ConsoleViewContentType.USER_INPUT` isn't doing what I expected it to.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ lazy val commonSettings = Seq(
 )
 
 val guava = "com.google.guava" % "guava" % "23.6-jre"
-val scalaTest = "org.scalatest" % "scalatest_2.12" % "3.0.4" % "test"
+val scalaTest = "org.scalatest" %% "scalatest" % "3.0.4" % Test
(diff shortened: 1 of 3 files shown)


rikvdkleij:
Thanks!

davips:
Why isn't the REPL behavior identical to GHCi? Isn't it GHCi?

rikvdkleij:
@davips Because it's runs inside IntelliJ.

PR #289, opened 2018-07-05
Accept lowercase pragmas too

ethercrow:
It's nice to be able to write ```{-# language Blah #-}``` without pressing Shift all the time.

### DIFF
--- a/src/main/scala/intellij/haskell/editor/HaskellCompletionContributor.scala
+++ b/src/main/scala/intellij/haskell/editor/HaskellCompletionContributor.scala
@@ -71,9 +71,10 @@ class HaskellCompletionContributor extends CompletionContributor {
   private final val SpecialReservedIds = Stream("safe", "unsafe")
   private final val PragmaIds = Stream("{-#", "#-}")
   private final val Ann = "ANN"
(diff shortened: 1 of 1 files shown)


rikvdkleij:
Thanks for your effort! 

> It's nice to be able to write {-# language Blah #-} without pressing Shift all the time.

But I do not see the relevance of this change. The constants are defined there for code completion. So it's not necessary to hold the Shift button all the time. Besides there is also a `Code Completion` setting in IntelliJ to enable no case sensitive completion.

dryewo:
Thanks @ethercrow for the effort.

I confirm the existence of the problem:
image

When I use lowercase `language`, Cmd-clicking to go to definition in this file does not work.
If I change pragmas to `LANGUAGE`, it starts working immediately.

rikvdkleij:
Because of no further activity I will close this REPL.

dryewo:
@rikvdkleij do you expect this problem to be addressed in the foreseeable future?

rikvdkleij:
@dryewo I will change the parser so it also accepts lowercase version of `LANGUAGE`

PR #302, opened 2018-09-02
fix(stack version): Correctly detect Stack version

### LINKED ISSUE #299: Stack version should be > 1.7.0
For some [Windows reasons](https://github.com/commercialhaskell/stack/issues/3797) I built and installed Stack from master branch (problems with bash coloring symbols being wrongly displayed in terminal) instead of the good ol' Windows installer coming from https://get.haskellstack.org/stable/windows-x86_64-installer.exe.

The problem being, the current Windows installer is 1.7.xxx while master branch is 1.10.0.

And yet, when I try to import a stack project into Intellij, in the Stack SDK window, when I try to add the path to my built 1.10.0 Stack exe, I get this error:

> 
> Select path to Stack binary
> Stack version should be > 1.7.0

I think there's something wrong with the Stack version being detected 😸

Sir4ur0n:
Closes #299 

This is my very first Scala code so please feel free to any criticism 😄

### DIFF
--- a/src/main/scala/intellij/haskell/sdk/HaskellSdkType.scala
+++ b/src/main/scala/intellij/haskell/sdk/HaskellSdkType.scala
@@ -25,6 +25,7 @@ import com.intellij.openapi.projectRoots.impl.SdkConfigurationUtil
 import com.intellij.openapi.roots.OrderRootType
 import com.intellij.openapi.util.SystemInfo
 import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.util.text.VersionComparatorUtil
 import intellij.haskell.external.execution.CommandLine
 import intellij.haskell.util.{HaskellFileUtil, HaskellProjectUtil}
 import intellij.haskell.{HaskellIcons, HaskellNotificationGroup}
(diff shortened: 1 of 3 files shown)


rikvdkleij:
Thanks! But this solution does not work in case minimum version is `1.7.1` and current version is `2.8.0`.

I had in mind to use method of IntelliJ SDK, for example `VersionComparatorUtil` of `com.intellij.util.text`

Sir4ur0n:
Spot on! My mistake.

I used `VersionComparatorUtil` which seems to do a perfect job for the use case you mentioned.

rikvdkleij:
Okay but now you have changed the behavior in case there is no current version which normally can not happen. Or is that intentionally?

Sir4ur0n:
This is definitely not my proudest moment, sorry 😢 

Yet another try, this time with tests to feel safer. Unlike your 2 other tests I used another syntax which feels more fluent for unit tests, tell me if it's not OK and I'll reuse the same syntax as the other 2.

By the way it looks like HLintSpec test doesn't pass (NPE), is that normal?

rikvdkleij:
> This is definitely not my proudest moment, sorry 😢

No problem, it's super that you are contributing to the project!

> Yet another try, this time with tests to feel safer. Unlike your 2 other tests I used another syntax which feels more fluent for unit tests, tell me if it's not OK and I'll reuse the same syntax as the other 2.

It's okay.

> By the way it looks like HLintSpec test doesn't pass (NPE), is that normal?

No, I did not run them for some while so did not notice that. They have to be fixed 😄

rikvdkleij:
Thanks!

Issue #315, opened 2018-10-03
Consider using OS-specific directories instead of ~/.intellij-haskell


liff:
It looks like intellij-haskell creates and uses `~/.intellij-haskell` as some sort of cache.

Would you consider using OS-specific cache directories instead, perhaps with the aid of [directories-jvm](https://github.com/soc/directories-jvm)?

Another option would be to use IDEA’s system directory but I guess that was not chosen for a reason.

rikvdkleij:
What is the problem of using `~/.intellij-haskell` as directory to store data?

Btw, it's not a cache, it is persistent data.

liff:
It’s not a problem, just a matter of convention. Since there are dedicated places for this sort of thing, why not use them?

liff:
Closing since done.

PR #316, opened 2018-10-04
Use OS-specific directory for peristent data

liff:
Implements #315. This will prefer and keep using an existing `~/.intellij-haskell` so that existing installations will not be affected. First run in a clean system will choose `cacheDir` using [soc/directories-jvm](https://github.com/soc/directories-jvm#basedirectories).

I can add a note to README if you think it’s worth mentioning?


rikvdkleij:
Thanks!

> I can add a note to README if you think it’s worth mentioning?

Yes, that would be helpful.

rikvdkleij:
On a second thought, maybe it's better to change it a way that the `.intellij-haskell` solution is no longer used. Otherwise we never get rid of it 😄 

I will ask opinions on the Gitter channel.

liff:
Yeah, I suppose the only inconvenience caused by changing the directory is that it’ll take a while to open a project since sources need to be re-downloaded and tools re-built?

rikvdkleij:
Thanks! I will change it so old directory is not used anymore.

PR #333, opened 2018-11-28
New logo for hs and cabal

ice1000:
IntelliJ has built-in retina screen support, you just need to provide `foo.png` and `foo@2x.png` in the same directory and let the `IconLoader` to automatically determine the correct icon to use.

New icons fit the newest IntelliJ's icon style.


rikvdkleij:
Thanks!

develop7:
Looks great!

PR #358, opened 2018-12-14
Support string literal language injection

### LINKED ISSUE #272: Multiline strings aren't highlighted
If you use`\` to create a multiline string, it's not highlighted like a regular string.

If you copy and paste the string, it's highlighted correctly.

Example: 
screen shot 2018-05-02 at 11 18 28 am

ice1000:
And fix #272 by the way.

### DIFF
--- a/gen/intellij/haskell/psi/HaskellTextLiteral.java
+++ b/gen/intellij/haskell/psi/HaskellTextLiteral.java
@@ -1,6 +1,8 @@
 // This is a generated file. Not intended for manual editing.
 package intellij.haskell.psi;
 
-public interface HaskellTextLiteral extends HaskellStringLiteralElement {
+import com.intellij.psi.PsiLanguageInjectionHost;
+
+public interface HaskellTextLiteral extends HaskellStringLiteralElement, PsiLanguageInjectionHost {
 
 }
--- a/intellij-haskell/META-INF/plugin.xml
+++ b/intellij-haskell/META-INF/plugin.xml
@@ -291,7 +291,7 @@
         <externalAnnotator language="Haskell"
                            implementationClass="intellij.haskell.annotator.HaskellAnnotator"/>
         <annotator language="Haskell"
-                   implementationClass="intellij.haskell.highlighter.HaskellSoftKeywordsAnnotator"/>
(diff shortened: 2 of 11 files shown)


ice1000:
[intellij-haskell.zip](https://github.com/rikvdkleij/intellij-haskell/files/2678253/intellij-haskell.zip)

ice1000:
DO NOT MERGE.
Some local changes cannot be pushed due to my PC's internet connection problem.
Please wait patiently.

ice1000:
Writing down some notes:

https://github.com/covscript/covscript-intellij/blob/master/src/org/covscript/lang/psi/cov-manipulators.kt
https://github.com/covscript/covscript-intellij/blob/master/src/org/covscript/lang/psi/impl/cov-psi-mixins.kt

rikvdkleij:
Thanks!

rikvdkleij:
@ice1000 I see that #272 is fixed but what else gives this change as feature to the user?

ice1000:
screen

Do you like what you see?

ice1000:
I've just found a bug. Will fix tonight

ice1000:
> Does it also work for SQL

Yes

> with code completion of table and field names

No :(

rikvdkleij:
It looks like this change broke the feature that inserts automatically closing `"` after entering starting `"`.

ice1000:
How?
Is there some old codes about that feature? I may take a look.

rikvdkleij:
The implementation is minimal, see `HaskellQuoteHandler`. I already added `HS_TEXT_LITERAL` but that does not help.

rikvdkleij:
Btw, I checked out revision before this change and there it still works.

ice1000:
Writing down some notes:

https://intellij-support.jetbrains.com/hc/en-us/community/posts/205418330-Quote-handler-in-template-language
https://intellij-support.jetbrains.com/hc/en-us/community/posts/206115129-Inserting-Closing-Quote-and-Bracket

rikvdkleij:
> You can see that it only accepts tokens, not Psi element types

You mean that the String Literal Injection feature changed the "behavior" of `"` so it's not a token anymore at that location?

ice1000:
No, I still don't know why is that happening. 😅

PR #387, opened 2019-02-17
Implement new parser, close #382, close #332

ice1000:
No, description, provided.

But please do add more tests to ensure that originally-accepted Haskell pragmas are still accepted in this new version.

PR #389, opened 2019-02-17
Trying to allow invalid syntax as "bad character" instead of crashing…

ice1000:
… the editor


rikvdkleij:
Thanks again!

ice1000:
I've reproduced the issue introduced in #387 and now I can't reproduce.

rikvdkleij:
Hmm, but `OPTIONS_GHC` is not used in bnf?

ice1000:
Yes. Is that something important? Now you can put anything seem-to-valid in the pragma (like an identifier `OPTIONS_GHC`, option `--OhMyGod`, or a string/char lit `" Spaces Around "`).

ice1000:
> Hmm, but `OPTIONS_GHC` is not used in bnf?

I've just realized that this is not only an old tokentype but a lexer state. I don't think it has something to do with the bnf.

rikvdkleij:
> Yes. Is that something important? Now you can put anything seem-to-valid in the pragma (like an identifier `OPTIONS_GHC`, option `--OhMyGod`, or a string/char lit `" Spaces Around "`).

No, I just learned something :smile: 
Normally I have to set it in the bnf otherwise the compiler complains.

I can confirm that is now works! Great!

Only the syntax highlighter does not highlight `LANGUAGE` anymore. 

I will fix other features like intention to add language pragma.

ice1000:
@rikvdkleij we're getting this parser for free.


I'm cleaning up the bnf and will PR again very soon.

rikvdkleij:
@ice1000  Super!

PR #394, opened 2019-02-24
fix(live template): Fix live template default context

### LINKED ISSUE #391: Live templates have a bad context
There are a few default live templates (e.g. `data`, `datag`) which have an invalid context. They have `Haskell / Global definition` enabled but it never prompts the template:
image

However by enabling `Other` it works:
image
image
image

The default context should be fixed for all applicable live templates.

Also, it may be cool to rename `Other` to something more explicit?

Sir4ur0n:
Closes #391.


rikvdkleij:
Thanks for helping!

PR #416, opened 2019-04-14
Move pluginIcon.xml

ice1000:
I'm sorry, it was in the wrong place.
2019-04-14 18-53-03 的屏幕截图
It now works.


rikvdkleij:
Thanks!

PR #424, opened 2019-05-23
Layout lexer: getting something very basic to work

ice1000:
I have a `SimpleLayout.hs`, which uses a `let` expression (who has got a layout inside of it), whose AST is also included in the PR (as test data). You can easily see the syntax tree -- it's awesome.
Also, to enable AST-dumping-based testing, I've implemented `toString` for some AST nodes so the printed AST does not change.

This is a ground PR for a large refactoring of the syntax definition. Any questions are welcomed, both here and on the gitter channel.

### DIFF
--- a/src/main/scala/intellij/haskell/psi/HaskellLayoutLexer.scala
+++ b/src/main/scala/intellij/haskell/psi/HaskellLayoutLexer.scala
@@ -47,6 +47,7 @@ class HaskellLayoutLexer(private val lexer: Lexer,
   private def currentToken = tokens.lift(currentTokenIndex)
 
   def start(buffer: CharSequence, startOffset: Int, endOffset: Int, initialState: Int) {
+    if (tokens.nonEmpty) return
     require(startOffset == 0, "does not support incremental lexing: startOffset must be 0")
     require(initialState == 0, "does not support incremental lexing: initialState must be 0")
 
@@ -129,12 +130,13 @@ class HaskellLayoutLexer(private val lexer: Lexer,
     val indentStack = new IndentStack()
     indentStack.push(-1) // top-level is an implicit section
 
-    for (token <- tokens) {
+    while (i < tokens.size) {
+      val token = tokens(i)
 
       state match {
         case WaitingForLayout =>
           if (token.isCode && token.column > indentStack.peek()) {
-            tokens.insert(i, virtualToken(layoutStart, tokens(if (i <= 0) 0 else i - 1)))
+            tokens.insert(i, virtualToken(layoutStart, tokens(i)))
             i += 1
             indentStack.push(token.column)
             if (!token.elementType.exists(layoutCreatingTokens.contains)) {
--- a/src/main/scala/intellij/haskell/psi/HaskellLayoutLexer.scala
+++ b/src/main/scala/intellij/haskell/psi/HaskellLayoutLexer.scala
@@ -79,6 +79,7 @@ class HaskellLayoutLexer(private val lexer: Lexer,
     var currentColumn = 0
 
     @tailrec
+    @inline
     def doIt(line: Line): Unit = {
       val token = Token(Option(lexer.getTokenType), lexer.getTokenStart, lexer.getTokenEnd, currentColumn, line)
       tokens += token
@@ -90,13 +91,16 @@ class HaskellLayoutLexer(private val lexer: Lexer,
       currentColumn += token.end - token.start
 
       if (!token.isEOF) {
-        if (token.elementType.contains(endOfLine)) {
+        val nextLine = if (token.elementType.contains(endOfLine)) {
           currentColumn = 0
+          Line()
+        } else {
+          line
         }
 
         lexer.advance()
 
-        doIt(Line())
+        doIt(nextLine)
       }
     }
 
--- a/src/main/scala/intellij/haskell/psi/HaskellLayoutLexer.scala
+++ b/src/main/scala/intellij/haskell/psi/HaskellLayoutLexer.scala
@@ -36,12 +36,12 @@ class HaskellLayoutLexer(private val lexer: Lexer,
     override def toString = s"${elementType.toString} ($start, $end)"
 
     val isEOF: Boolean = elementType.isEmpty
-    val isCode: Boolean = elementType.exists(et => !nonCodeTokens.contains(et)) && !isEOF
+    def isCode: Boolean = elementType.exists(et => !nonCodeTokens.contains(et)) && !isEOF
 
-    val isNextLayoutLine: Boolean = isCode && line.columnWhereCodeStarts.contains(column)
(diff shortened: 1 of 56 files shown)


rikvdkleij:
Thanks!

src/main/scala/intellij/haskell/psi/HaskellLayoutLexer.scala
```diff
@@ -36,17 +36,18 @@ class HaskellLayoutLexer(private val lexer: Lexer,
     override def toString = s"${elementType.toString} ($start, $end)"
 
     val isEOF: Boolean = elementType.isEmpty
-    val isCode: Boolean = elementType.exists(et => !nonCodeTokens.contains(et)) && !isEOF
+    def isCode: Boolean = elementType.exists(et => !nonCodeTokens.contains(et)) && !isEOF
```
rikvdkleij: `isCode` can be `val` because depends on `val`s

src/main/scala/intellij/haskell/psi/HaskellLayoutLexer.scala
```diff
@@ -36,17 +36,18 @@ class HaskellLayoutLexer(private val lexer: Lexer,
...
 
-    val isNextLayoutLine: Boolean = isCode && line.columnWhereCodeStarts.contains(column)
+    def isNextLayoutLine: Boolean = isCode && line.columnWhereCodeStarts.contains(column)
```
rikvdkleij: `isNextLayoutLine` can be also be `va`l if we decouple `Line` from `Token`. Much nicer.

src/main/scala/intellij/haskell/psi/HaskellLayoutLexer.scala
```diff
@@ -36,17 +36,18 @@ class HaskellLayoutLexer(private val lexer: Lexer,
...
   }
 
-  private val tokens = new ArrayBuffer[Token](40)
+  private val tokens = new ArrayBuffer[Token](100)
   private var currentTokenIndex = 0
 
   private def currentToken = tokens.lift(currentTokenIndex)
 
   def start(buffer: CharSequence, startOffset: Int, endOffset: Int, initialState: Int) {
+    if (tokens.nonEmpty) return
```
rikvdkleij: Hmm, why this check if tokens are not empty?

rikvdkleij:
When I have this code
```
module LayoutTest where

test =
  let x = 1
      y = 2
   in x
```

In the PSI-tree I see the distinction between `let` and `in`. So next step is to make distinction between `x = 1` and `y = 1`, if I'm right :smile: 

This first step is really great!

Btw, I also see that "lost" opening layout brace not anymore.

src/main/scala/intellij/haskell/psi/HaskellLayoutLexer.scala
```diff
@@ -79,24 +80,28 @@ class HaskellLayoutLexer(private val lexer: Lexer,
     var currentColumn = 0
 
     @tailrec
+    @inline
```
rikvdkleij: Why did you inline here?

rikvdkleij:
This code give parser error at the end:
```
main :: IO()
main = do
  let s = "s"
  print s
```

src/main/scala/intellij/haskell/psi/HaskellLayoutLexer.scala
ice1000: Yes. This was a mistake.

src/main/scala/intellij/haskell/psi/HaskellLayoutLexer.scala
ice1000: It is. I agree, but before spotting the other mistranslations, I tried to fix the lexer by modifying these codes and I forgot to change them back

src/main/scala/intellij/haskell/psi/HaskellLayoutLexer.scala
ice1000: Oops, please delete it! The lexer was invoked twice during debugging and I didn't want my breakpoints to stop twice.

ice1000:
> In the PSI-tree I see the distinction between let and in. So next step is to make distinction between x = 1 and y = 1, if I'm right 😄


Yes, Yes, Yes, Yes!

ice1000:
> This code give parser error at the end:

Because `DO` is not yet a `layoutStartingToken`

src/main/scala/intellij/haskell/psi/HaskellLayoutLexer.scala
ice1000: Feel free to delete it. This was unnecessary.

src/main/scala/intellij/haskell/psi/HaskellLayoutLexer.scala
```diff
@@ -126,12 +130,13 @@ class HaskellLayoutLexer(private val lexer: Lexer,
     val indentStack = new IndentStack()
     indentStack.push(-1) // top-level is an implicit section
 
-    for (token <- tokens) {
+    while (i < tokens.size) {
+      val token = tokens(i)
```
rikvdkleij: Can this not back to the previous `for` loop?

It’s weird that `while` is over size with `i`and within while `i` is changed.

ice1000: > It’s weird that while is over size with iand within while i is changed.

`i` is tweaked inside of the loop (and it's intentional, like we're ignoring some tokens), and we're mutating `tokens` while traversing it.
Can't use `for`.

ice1000: I spent two days to realize this mistake

rikvdkleij: Ok but it makes the logic very very hard to understand

rikvdkleij: I have no idea how this code behaves with respect to `i`

ice1000: > Ok but it makes the logic very very hard to understand

It is. The original code was very very very imperative

ice1000: > I have no idea how this code behaves with respect to `i`

`i` was assigned many times, as you can see

rikvdkleij: Yes, I see but it comes back in the `while` and then?

rikvdkleij: I would like to rewrite (further) it but I have to understand what is going on.

ice1000: > I would like to rewrite (further) it but I have to understand what is going on.

Yes, now we have tests to ensure the lexer behavior.

PR #426, opened 2019-05-24
Address comments, support `do` notation

ice1000:
See the added test.

Note: I'm struggling with the generated code. I guessed you're running "code cleanup" before committing them, so did I. However, some of them are 2-spaces-indented, some are 4-spaces. I'm lost -- I regenerate the code and got most of them changed.
@rikvdkleij there should be a solution.

### DIFF
--- a/gen/intellij/haskell/parser/HaskellParser.java
+++ b/gen/intellij/haskell/parser/HaskellParser.java
@@ -2298,6 +2298,67 @@ public class HaskellParser implements PsiParser, LightPsiParser {
     }
 
     /* ********************************************************** */
+    // (let_layout | expression) SEMICOLON?
+    static boolean do_clause(PsiBuilder b, int l) {
+        if (!recursion_guard_(b, l, "do_clause")) return false;
+        boolean r;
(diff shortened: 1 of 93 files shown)


ice1000:
I'll recommend setting a standard, like using `.editorconfig`.

rikvdkleij:
> I'll recommend setting a standard, like using .editorconfig.

I would suggest to leave the code as is generated. If that is possible with `.editorconfig` then that is fine.

rikvdkleij:
> I'm lost -- I regenerate the code and got most of them changed.
@rikvdkleij there should be a solution.

That is because I always format and optimize imports when I commit. But, yes, that is not a good practice for the generated code.

rikvdkleij:
Thanks again!

ice1000:
> I would suggest to leave the code as is generated. If that is possible with .editorconfig then that is fine.

The generated code's format depends on your IDE settings -- the plugin at least reads your code style configuration.
With `.editorconfig`, the code style will be enforced to the specified one.

Issue #417, opened 2019-04-16
Intellij IDEA fails to upgrade to beta 46


mouse07410:
On one machine the update succeeded, and `beta 46` seems much better than the previous versions.

On all the other machines, this plugin is stuck at version `beta 43`, and I seem unable to force IDEA to update it. 

If I don't want to rebuild the plugin from the source - where can I download it from?

Also, could the problem be related to the possible name mismatch? Here *module-name=*`IntelliJ Haskell`, in other places *module-name=*`IntelliJ-Haskell`
https://github.com/rikvdkleij/intellij-haskell/blob/69199f33f1d555ba8ac6fee70a65b585441c4cb2/intellij-haskell/intellij-haskell.iml#L9

mouse07410:
It looks like changing the "Custom plugin repositories" to `https://plugins.jetbrains.com/plugins/alpha` (dropping the tail `/8258`) resolved this problem, and made the plugin update to `beta 46` immediately.

PR #432, opened 2019-06-13
Update alpha repo url

AllanWang:
As per https://github.com/rikvdkleij/intellij-haskell/issues/417, it looks like the `/8258` part is no longer necessary. Builds get stuck on `beta43` otherwise


rikvdkleij:
Thanks!

Issue #436, opened 2019-06-16
Quick documentation for (+) shows documentation for (++)


AlainODea:
If I have the following code:
```haskell
f x y = x + y
```

And I place my caret before the `+` and click View > Quick Documentation (or hit F1 on macOS or Ctrl-Q on Windows/Linux) I get the following popup:

> ```
> (++) :: [a] -> [a] -> [a] base Prelude
> ```
> ----- 
> ```
> Append two lists, i.e.,
> 
> [x1, ..., xm] ++ [y1, ..., yn] == [x1, ..., xm, y1, ..., yn]
> [x1, ..., xm] ++ [y1, ...] == [x1, ..., xm, y1, ...]
> 
> If the first list is not finite, the result is the first list.  
> ```
> 
> -----
> GHC.Num    -- No type info available
> 
> -----

Here's a screenshot of the behaviour:
![Quick documentation for (+) shows documentation for (++)](https://user-images.githubusercontent.com/16341/59566524-1407be00-903c-11e9-9613-b158e605be89.png)

I tried to follow HaskellDocumentationProvider's use of TypeInfoComponent and StringUtils. I couldn't find the issue in StringUtils so I suspect it is somewhere down the TypeInfoComponent rabbit hole.

I've gotten the development setup working following your excellent [CONTRIBUTING.md](https://github.com/rikvdkleij/intellij-haskell/blob/master/CONTRIBUTING.md), so I will debug this and hopefully have a PR to fix it fairly soon.

AlainODea:
The problem is something in how HoogleComponent encodes the arguments for the hoogle command. It seems like searching for `+` literally in hoogle on the CLI also does not work.

Literally running this command brings up the documentation for (++):
```shell
/Users/alain_odea/Library/Caches/com.github.rikvdkleij.intellij-haskell/lts-13/bin/hoogle --database=/Users/alain_odea/Library/Caches/com.github.rikvdkleij.intellij-haskell/finding-success-and-failure/hoogle -i + +GHC.Num +Prelude
```

It looks like the bug is in hoogle. I'll see how to encode + to work around this.

AlainODea:
Show Type (Option-Command-P) correctly shows `(+) :: Num a => a -> a -> a`.

[...]

PR #437, opened 2019-06-16
Have (+) show correct quick documentation (none). Fix issue #436.

AlainODea:
Addressing #436 by adding is:exact to the arguments sent to the hoogle CLI to avoid showing the documentation for (++) when requesting the documentation for (+).

The popup now shows:
> ```
> No documentation found
> ```
> -----
> GHC.Num    -- No type info available
> 
> -----
Quick documentation for (+) shows no documentation found as expected

### DIFF
--- a/src/main/scala/intellij/haskell/external/component/HoogleComponent.scala
+++ b/src/main/scala/intellij/haskell/external/component/HoogleComponent.scala
@@ -102,7 +102,7 @@ object HoogleComponent {
 
     ProgressManager.checkCanceled()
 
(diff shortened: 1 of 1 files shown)


rikvdkleij:
Thanks!

AlainODea:
My pleasure. Thank you for this awesome IDE for Haskell!

rikvdkleij:
My pleasure :smile: 

Btw, locally I already changed the hoogle call because doing the call with `+ moduleName` does not always work. So I changed module name to package name which gives better results.

PR #440, opened 2019-06-18
Fix / improve English in various user / log messages.

### LINKED ISSUE #439: Improve error / log messages
Hi. Thanks again for your hard work on such a great plugin.

I've noticed some of the error / log / IDE user messages feel a bit weird or unfriendly (to a native speaker at least). Some examples...

Nothing wrong - but it sounds a bit odd (IMO)

> _Can not load file in REPL because could not determine path for file `${actionContext.psiFile.getName}`. File exists only in memory"_

Again - fine, but we could maybe make a clearer message here..

> _Busy with preloading..._

_Busy with x_ sounds weird - just _Busy x_ sounds better

### Errors / Typos
e.g. 
> _Can not add package $p as module because it's absolute file path ${packagePath.getAbsolutePath} does not exist._

That should be _its_

> _Busy with preloading stack component info cache_

Should probably be _Stack_

Hardly urgent, of course, but always nice to have friendly messages!  I can try to push some fixes if that helps :up:

declension:
* Some contractions for friendliness / readability (e.g. _can not_ -> _can't_)
 * Some fixes for consistency in similar message 
 * Fix a few typos
 * Rewrite (slightly) a few of the longer / complicated ones to try to clarify the problem

Please revert any that don't work for you of course (or I can)

This closes #439.

### DIFF
--- a/src/main/scala/intellij/haskell/action/ghci/LoadModuleAction.scala
+++ b/src/main/scala/intellij/haskell/action/ghci/LoadModuleAction.scala
@@ -19,7 +19,7 @@ class LoadModuleAction extends AnAction {
     } yield {
       HaskellFileUtil.getAbsolutePath(actionContext.psiFile) match {
         case Some(filePath) => consoleView.executeCommand(s":load $filePath", addToHistory = false)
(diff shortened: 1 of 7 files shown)


rikvdkleij:
Great! Thanks!

PR #558, opened 2020-05-13
Can't find cabal file

i-walker:
In some of our projects, we had issues opening haskell projects that have the following structure:

zasi

```
.
├── .idea // <- contains the .iml file here
├── Setup.hs
├── dir1
│   └── task1
│       ├── Task.hs
│       ├── Test.hs
│       └── task.md
├── haske89.cabal
├── stack.yaml
```
And whenever `HaskellProjectUtil.getModuleDir` in https://github.com/i-walker/intellij-haskell/blob/382179707700d835346ad271cc1a4f67d69629f5/src/main/scala/intellij/haskell/external/repl/StackReplsManager.scala#L72. the list will be empty as the Path will be the `.idea` directory. 
The project gets loaded as is, but the user experience is not nice seeing this every time.

I am open for Feedback.

### DIFF
--- a/src/main/scala/intellij/haskell/util/HaskellProjectUtil.scala
+++ b/src/main/scala/intellij/haskell/util/HaskellProjectUtil.scala
@@ -25,6 +25,7 @@ import com.intellij.openapi.util.io.FileUtil
 import com.intellij.openapi.vfs.{VfsUtilCore, VirtualFile}
 import com.intellij.psi.search.GlobalSearchScope
 import com.intellij.psi.{PsiElement, PsiFile}
+import com.intellij.util.PathUtilRt
 import intellij.haskell.GlobalInfo
 import intellij.haskell.external.component.HaskellComponentsManager
 import intellij.haskell.module.HaskellModuleType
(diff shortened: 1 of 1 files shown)


rikvdkleij:
Thanks for the PR.

To understand the issue: Why is the `.iml` file in the `.idea` directory? AFAIK it's always in the root of the project.

i-walker:
That is partially correct. 
In some cases, when creating a project the `.iml` file gets saved ini the idea folder with.the rest of the ide related xml files and etc.

i-walker:
That might have changed since 2015, which was the last time this file was updated

src/main/scala/intellij/haskell/util/HaskellProjectUtil.scala
```diff
@@ -78,7 +79,12 @@ object HaskellProjectUtil {
   }
 
   def getModuleDir(module: Module): File = {
-    new File(ModuleUtilCore.getModuleDirPath(module))
+    val path = ModuleUtilCore.getModuleDirPath(module)
+    new File(
+      if (path.endsWith(".idea"))
```
rikvdkleij, line 84:
```suggestion
      if (path == ".idea")
```

rikvdkleij: I think using `==` is more to-the-point :smile:

i-walker: I've run some tests and that will fail for the issue I am adressing

i-walker: the path here is "<User/....../workspace/..../.idea>" not the name of the directory

rikvdkleij: Ok, I see

rikvdkleij:
Thanks!

PR #565, opened 2020-05-23
Update a few methods for 2020.1

i-walker:
- regarding test config the Id has to be overridden.
- annotation methods
- live template registration


rikvdkleij:
Thanks again!

PR #596, opened 2020-08-02
Use ANSI escape decoder for command-line output

### LINKED ISSUE #585: Use ANSI escape sequence processor in the initial tool setup progress bar
As I noted in #584, during the initial tool setup (at least in 1.0.0.-beta74 build on Windows, IDEA #IU-203.833), the following image is shown in the IDE progress window:

image

The latest progress line always starts with `[0m`, which, I believe, is an ANSI escape sequence fragment.

I believe that IntelliJ-Haskell could process this escape sequence and provide more accurate text information to the user, i.e. not show this `[0m`. IDEA code base has the corresponding utility functions available.

ForNeVeR:
This will use IDEA-backed ANSI decoder to decode the escape sequences, so `[0m` won't be shown to the user.

Closes #585.


rikvdkleij:
@ForNeVeR Thanks!

Issue #623, opened 2020-11-02
External tool output parsing breaks file encoding (UTF-8)


sgt:
For example, when reformatting Haskell code which contains Unicode symbols (.e. emoticons) using either Ormolu or Stylish Haskell, the Unicode symbols return garbled.
The tools themselves do not break the encoding, my guess the culprit is the code which parses tools' output and loads it back into the editor.

rikvdkleij:
Can you give an example?

I can't reproduce this issue.

rikvdkleij:
![image](https://user-images.githubusercontent.com/509315/97908846-8c614300-1d47-11eb-9683-a5737576cecc.png)

sgt:
![image](https://user-images.githubusercontent.com/5119/97917473-92feb300-1d65-11eb-83af-5e4a11c5071c.png)

After pressing Ctrl-Alt-L:

![image](https://user-images.githubusercontent.com/5119/97917573-b590cc00-1d65-11eb-990e-bf02a8fb26fd.png)

I'm on Windows. Finally got the plugin to compile, will try to investigate further if you can't manage to reproduce it on your side.

rikvdkleij:
@sgt Again, your help is really appreciated!

PR #624, opened 2020-11-03
Making command line processes charset aware. Fix issue #623

sgt:


sgt:
This seems to have fixed it. Command line shell that processes a file must run with locale/codepage matching the file's encoding.

rikvdkleij:
Thanks!

Would be nicer if the charset would be some global setting instead of managing this per file but okay. Great you fixed this!

sgt:
I'm no expert on Intellij internals, but it seems to me that the encoding is a property of a specific file (UI-wise shown on the status bar in the bottom right corner). You may hypothetically have files using different encodings in the same project.

Anyway, thanks for making this great plugin!

Issue #638, opened 2021-02-06
No documentation in code completion


fantostisch:
When using IntelliJ with other languages it is possible to press CTRL+Q to open the documentation for suggested completions. Example for Scala:
![scala](https://user-images.githubusercontent.com/3309700/107123397-9c844380-689d-11eb-8243-cd95bb744adc.png)

This does not seem to work with this plugin:
![intellisense](https://user-images.githubusercontent.com/3309700/107123405-aa39c900-689d-11eb-9100-8a4fde5db9e1.png)

Even though the documentation is shown when the full function is written:
![documentation](https://user-images.githubusercontent.com/3309700/107123429-c63d6a80-689d-11eb-8009-4b1c089e2066.png)

rikvdkleij:
Sorry, I don’t understand the issue. The documentation is shown in latest screenshot. Maybe the difference in layout confuses.

fantostisch:
It would be nice if the documentation would be shown of what is suggested in the window. Someting like this:
![intellisense2](https://user-images.githubusercontent.com/3309700/107151374-8be9d100-6962-11eb-9c0d-6cbc7c6bbc7d.png)

I will try to implement this

fantostisch:
It looks like we have to add a PsiElement to every LookupElement. Doing someting like `createLookupElement(_).withPsiElement(...)` in
https://github.com/rikvdkleij/intellij-haskell/blob/64ff54e04ec751edcd8f184a79c5eff4a775ae85/src/main/scala/intellij/haskell/editor/HaskellCompletionContributor.scala#L413
should work, but I have no idea were to get the right PsiElement from.
This PsiElement will then be send to
https://github.com/rikvdkleij/intellij-haskell/blob/8c3d788e80606c8fc598214ded5a5c1a03bff209/src/main/scala/intellij/haskell/external/component/HaskellDocumentationProvider.scala#L60
when pressing CTRL+Q.

rikvdkleij:
I guess you can use this to find the element by using the module identifier:
http://github.com/rikvdkleij/intellij-haskell/blob/4f561c252738d4359f6fe06f819ee154aebb3c6c/src/main/scala/intellij/haskell/navigation/HaskellReference.scala#L205-L205

Hopefully it doesn't affect the responsiveness.

fantostisch:
When importing TVar using
```hs
import qualified Control.Concurrent.STM.TVar as TVar
```
[...]

PR #645, opened 2021-02-20
Documentation in code completion

fantostisch:
Fixes issue #638

Preview:
screenshot-doc

Using `originalElement` in

```scala
getQuickNavigateInfo(element: PsiElement, originalElement: PsiElement)
```
does not work because it seems to be `PsiWhiteSpace` when using code completion, so I changed it to always use `element`. I am not sure if this breaks other features.

Sometimes I can not navigate through the options using the arrow keys on my keyboard, only using my mouse. I don't know why this is happening.

The documention lookup is only done when pressing CTRL+Q so it should not affect the responsiveness.


fantostisch:
Hmm, this also tries to find an identifier for completions that are not modules like the completion for "where". Will try to fix this.

rikvdkleij:
@fantostisch thanks! I will take a look soon.

rikvdkleij:
@fantostisch Thanks!

PR #681, opened 2022-03-19
Add default sdk home path for windows

SimonIT:
It's the recommended path by stack https://docs.haskellstack.org/en/stable/install_and_upgrade/#path


rikvdkleij (approved):
Thanks!

